PR #16, opened 2018-04-12
Create safe_usecases.md

pragashj:
First version of the use case translated from the doc


ultrasaurus:
added a separate issue to add Adminstrator Bill of Rights to the repo https://github.com/cn-security/safe/issues/19 -- rather not have a link to an external Google doc :)

PR #40, opened 2018-07-30
Adding my identity to list

knowlengr:
Whew, no work to do here.


dshaw:
Landed in https://github.com/cn-security/safe/pull/50

PR #76, opened 2018-09-06
Add use case for meeting compliance requirements

lizrice:
I think we should add a use case about meeting compliance requirements

### DIFF
--- a/safe_usecases.md
+++ b/safe_usecases.md
@@ -100,7 +100,7 @@ Compliance Officer /Auditor
 
 * As a compliance officer, I can understand the effect of changes to audit logging policy that I am making.
 
-* As a compliance officer, I can configure my organization's resources to meet the requirements of relevant standards such as PII or HIPAA.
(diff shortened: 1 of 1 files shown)


ultrasaurus:
I like this addition.  Added to Friday's agenda for review.

safe_usecases.md
```diff
@@ -98,7 +98,9 @@ Compliance Officer /Auditor
 
 * As a compliance officer, I can set audit logging policy that controls what data gets collected for auditing purposes.
 
-* As a compliance office, I can understand the effect of changes to audit logging policy that I am making
+* As a compliance officer, I can understand the effect of changes to audit logging policy that I am making.
+
+* As a compliance officer, I can configure my organization's resources to meet the requirements of relevant standards such as [PCI](https://www.pcisecuritystandards.org/), [FedRAMP](https://www.fedramp.gov/) or [HIPAA](https://www.gpo.gov/fdsys/pkg/PLAW-104publ191/html/PLAW-104publ191.htm).
```
anweiss: I would maybe expand upon this a bit and include something along the lines of:

```
... and automate the development of attestation and assessment artifacts for meeting the requirements of relevant standards ...
```

lizrice: Love the idea of the requirement for assessment and attestation artifacts. I used the word "generate" rather than "automate" to be slightly less prescriptive (I think automation is a marvellous thing, but even a manual check-list would be a very helpful step, right?)

anweiss: Awesome! Yea, I like “generate” better as well.

PR #82, opened 2018-09-14
Add September meeting dates

dshaw:


duglin:
Would it not be easier to just say we meet weekly?

PR #102, opened 2018-11-08
update objective

ultrasaurus:
Now that we've been doing this for a while, we have a better idea how to describe it.  
I wanted to capture something a lot more active than "explore" 
In Nov 8 work session @pragashj @lizrice @dshaw and I came up with this


ultrasaurus:
Merging this PR, since @dshaw addressed feedback by merging the change into this one -- thanks!

PR #111, opened 2018-12-02
lazy consensus applies to any proposal, not just candidates

ultrasaurus:
small adjustment to wording, submitting as PR since it changes the meaning 
but I believe it is consistent with the rest of the document


ultrasaurus:
don't think these needs all three of you, but would be good if @rachelmyers @dshaw or @pragashj could review and merge if this matches your understanding

PR #113, opened 2018-12-02
description of quota operator

ultrasaurus:
addresses https://github.com/cn-security/safe/issues/32, I've annotated Quota Operator with a description based on comments by @ckemper67 in the issues

Sorry about the whitespace diffs... viewing the "rich diff" will show just the added text:

Since quota is often used for cost control, this may imply a different persona with financial, rather than an engineering background.

An important use of quota is to protect a service from abuse. By setting a quota, we can ensure that a single individual or group cannot bring down the service for everybody else (either intentionally or unintentionally). For example, services may lack sufficient protections (such as exponential backoff) and a simple quota enforcement in front of the service can reduce the impact of repeated request on the rest of the infrastructure.

### DIFF
--- a/safe_usecases.md
+++ b/safe_usecases.md
@@ -1,4 +1,4 @@
-Authors: stummidi@pivotal.io, rayc@google.com, pragashjj@gmail.com, ckemper@google.com 
+Authors: stummidi@pivotal.io, rayc@google.com, pragashjj@gmail.com, ckemper@google.com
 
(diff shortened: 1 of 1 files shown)


ultrasaurus:
cc Authors of the original doc: stummidi@pivotal.io, rayc@google.com, pragashjj@gmail.com, ckemper@google.com -- PTAL

Issue #66, opened 2018-08-30
Draft categories for SAFE Landscape


ultrasaurus:
We need to draft the categories that we'll use for grouping projects in the landscape.

This issue is to create an initial draft, and then we'll discuss at a WG meeting

lizrice:
@izgeri thanks for flagging that. For reference here are what I think are the [relevant meeting notes](https://docs.google.com/document/d/1vkUJcj6m8OzZorgODk_D1bchnByb4EXJLbKf5tcMSXA/edit#heading=h.p0wjx8ruop5j) from the Reference Architecture group

izgeri:
Thanks @lizrice. There are also conversations that have been happening in that [mailing list](https://lists.cncf.io/g/cncf-reference-architecture/topics), and I expect @mfdii and I will have a proposal in the next two weeks (he's away this week).

pragashj:
@evan2645, @sreetummidi please take a look at this

izgeri:
@dshaw do I need to be assigned this issue? I'm not sure how to do that in this repository.

izgeri:
@pragashj @lizrice @ultrasaurus @evan2645 @sreetummidi @dshaw 

[This is the Google document](https://docs.google.com/document/d/1--nMpThR_LyUUBN_7wVBC0L1IGzMAbY7kQkiPVR9sH0/edit?usp=sharing) with notes on what @mfdii and I have worked up so far. It includes an explanation of the purpose of our work as well as a proposed structure of what we have been calling the "Security Landscape".  I will be letting the Reference Architecture mailing list know about this draft as well, and that we will be discussing the draft in this group on Friday.

We are eager to get feedback, and the document is open for comments.

izgeri:
Task breakdown with tasks merged from #69 (I am unable to edit the original issue):

# Define Categories / Sub-categories
- [X] propose initial draft of security-focused landscape
- [X] discuss initial categorization proposal in SAFE WG
- [ ] solicit feedback from subject matter experts
- [ ] solicit feedback from CNCF projects
  - Security and Policy solutions
    - [ ] SPIFFE / SPIRE - Evan
    - [ ] Notary
    - [ ] TUF
    - [ ] OPA - Torin with updater
  - Projects with Security Concerns
    - [ ] gRPC
  - Platforms (consider sub-projects that can be used stand-alone, such as Envoy, Mixer):
    - [ ] Istio
    - [ ] Kubernetes
[...]

PR #122, opened 2019-01-10
landscape refactor

ultrasaurus:
This is not intended to change any of the content of @izgeri landscape PR, which is the result of a lot of great work of many SAFE WG members

In this PR, I made one commit, which is taking https://github.com/cn-security/safe/pull/119 verbatim and separating into 3 docs in a subfolder, then as a separate commit, I edited the language to fit more with README and separate docs which could be referenced from there


landscape/README.md
```diff
@@ -0,0 +1,36 @@
+## Goal
+
+The [SAFE roadmap](../roadmap.md) includes describing the landscape of
+cloud-native security. We evaluated categories in the CNCF Landscape and
```
izgeri: It would be good to link CNCF Landscape in its first appearance in the doc (ie here)

landscape/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Landscape Process
+
+- [X] Draft [categories](categories.md)
+- [ ] Determine approach to category mapping (see below)
+- [ ] Map cloud-native tools into categories (adjusting categories as needed)
+- [ ] Validate categories and landscape with review by makers and users of
+    cloud-native security solutions, as well as partner working groups
+
+A note on how the work of mapping tools into the sub-categories may proceed:
```
izgeri: Maybe make this into a subsection, that we can then link above (where it says "see below")

landscape/categories.md
```diff
@@ -0,0 +1,56 @@
...
+
+- **App definition and development**
+  - _Static Code Analysis_
+    - Inspecting code for OWASP vulnerabilities, pen testing, etc
+  - _Dependency analysis_
+    - Checking OS for vulnerabilities (updates available? Reports from https://www.us-cert.gov/ etc), image scanning
+    - Check for vulnerabilities in dependent libraries
+    - Check for maintenance of dependent projects => increased potential for vulnerabilities
+  - _Pipeline security testing_
+    - Tools that facilitate automated security testing in pipelines; eg for functional security tests of authn and authz,
```
izgeri: Should this category be renamed to be clearer about what it contains (right now, just tools to automate security testing)? Should we enlarge it to also include other tools to secure CI/CD pipelines? How does supply chain management fit in?

landscape/categories.md
```diff
@@ -0,0 +1,56 @@
...
+      - Firewalls, network segmentation, network policy
+  - _Service access_
+    - Service admittance & admission controllers
+      - Allowing new service instances to join existing services, and preventing insecure application instances from running.
+    - Storage access
+  - _Key and Certificate Management_
+    - Provisioning of access to required services for application instances, service accounts, orchestrator access to shared components, etc
+    - Provisioning of SSL certs for application instances
+
+- **Privacy**
```
izgeri: Where does data sovereignty fit into in this category? For example, keeping European data siloed or preventing data with different levels of classification from mixing (at rest or in motion)

landscape/categories.md
```diff
@@ -0,0 +1,56 @@
...
+  - _Storage Security_
+    - Data colocation and encryption at rest/motion
+    - Preventing insider access to protected data
+    - Monitoring for accidental data spills
+
+- **Provisioning**
+  - _Automation & Configuration Compliance_
+    - Compliance checkers, check platform configurations, verify private resources are not unexpectedly publicly accessible (eg S3 buckets), etc
+  - _Secure Container Registries_
+    - Image signing, etc
```
izgeri: Should be clarified to be clear that secure container registries contain signed/verified images
Image signing / attestation should be moved up into the App Definition and Development as part of the pipeline-related sub-categories

Revise sub-category to Trusted Compute and add Secure Container Registries, self-hosted package repositories, etc as examples

izgeri:
Noting that this is the updated PR connected to #66

landscape/categories.md
ericavonb: I was thinking about projects like this one:  [Open Cloud Integrity Technology](https://01.org/opencit)

landscape/categories.md
ultrasaurus: @dshaw and I thought data sovereignty was same as data location, so made that change

PR #148, opened 2019-04-15
light edits for repo rename

ultrasaurus:
addressed only the name of the group/repo

Note: there'll need to be additional edits to include reference to TOC, adjustments to process, etc.


caniszczyk:
LGTM

PR #188, opened 2019-05-30
Update categories.md

omerlh:
Added dynamic analysis and secret management categories. Rewrite static analysis - I don't like the term "OWASP vulnerabilities", it's unclear as there is no such thing. The OWASP top 10 is not strictly a list of vulnerabilities. Maybe common or popular vulnerabilities?


landscape/categories.md
```diff
@@ -6,7 +6,9 @@ sub-categories. The remaining text is a description of each sub-category.
 
 - **App definition and development**
   - _Static Code Analysis_
-    - Inspecting code for OWASP vulnerabilities, pen testing, etc
+    - Inspecting code for vulnerabilities, pen testing, etc
+  - _Dynamic Analysis_
+    - Checking application in runtime for various vulnerabilities
```
ultrasaurus: This seems like a good addition

landscape/categories.md
```diff
@@ -18,6 +20,8 @@ sub-categories. The remaining text is a description of each sub-category.
   - _Pipelines_
     - Tools that ensure a secure pipeline or workflow, for example,
     as applied to devops (CI/CD), supply chain, etc.
+  - _Secret Management_
+    - Tools that help to manage, encrypt or distribute secrets and other sensitive information.
```
ultrasaurus: I think this is the same as "Key and Certificate Management" (which I think is in the other section since it is needed across all parts of cloud technology, not just "App definition and development")

landscape/categories.md
```diff
@@ -6,7 +6,9 @@ sub-categories. The remaining text is a description of each sub-category.
 
 - **App definition and development**
   - _Static Code Analysis_
-    - Inspecting code for OWASP vulnerabilities, pen testing, etc
+    - Inspecting code for vulnerabilities, pen testing, etc
```
ultrasaurus: I think that OWASP was intended to be an example, but agree with dropping it since it can read as overly specific.  Thanks for reviewing this!

landscape/categories.md
omerlh: It's not necessarily the same I think, but maybe it can both go into more generic categories. Keys and certificates (at least IMO) are more areas of TLS/secure communication. Maybe secret distribution system? This is the name used by Istio I think.

ultrasaurus: Would be helpful to have some examples of open source projects that fall into this category vs Key and Certificate Management

omerlh: There are a few, projects like [Hashicorp Vault](https://www.vaultproject.io/), [Sealed Secrets](https://github.com/bitnami-labs/sealed-secrets) or [Kamus](https://kamus.soluto.io) (which is also proposed as a CNCF project). Vault can do also key management IIRC, and there is also [Lemur](https://github.com/Netflix/lemur).

Issue #166, opened 2019-05-20
[Assessment] in-toto


ultrasaurus:
Project Name: in-toto

Github URL: https://github.com/in-toto

Security Provider: yes

- [x] Identify team
   - [x] Project security lead: Santiago Torres Arias
   - [x] Lead Security Reviewer: Sarah Allen (@ultrasaurus)
   - [x] additional reviewers: Justin Cormack, Brandon Lum
- [x] Create slack channel (e.g. #sec-assess-intoto)
- [x] Project lead provides [draft document](https://docs.google.com/document/d/1O3O7eFCTZ_7lLZ9IhHSWFiXhjWf2gPuJbzZ7u0m-9uI/edit#heading=h.oykub2ep8ndw)
- [x] "Dumb question phase" Lead Security Reviewer asks clarifying questions 
- [x] Initial review
- [x] Project lead addresses comments/questions in [draft doc](https://docs.google.com/document/d/1O3O7eFCTZ_7lLZ9IhHSWFiXhjWf2gPuJbzZ7u0m-9uI/edit#)
- [x] Presentation & discussion - TODO Link recording, meeting minutes
- [x] Share draft findings with project - in progress
- [x] Assessment summary and doc checked in https://github.com/cncf/sig-security/tree/master/assessments/projects/in-toto
- [X] CNCF TOC presentation -- July 9, 2019 https://www.youtube.com/watch?v=7R6IfBFg2VA (40:00)

JustinCappos:
The in-toto team has reviewed the documentation and revised the documentation.  I believe we're ready for the assessment team to take the next step

ultrasaurus:
@JustinCappos added checkbox for "Project lead addresses comments/questions in draft doc" (we should add that to the template) -- doesn't block our finalizing the summary, but does block "Assessment summary and doc checked into..."

dshaw:
👍 https://github.com/cncf/sig-security/tree/master/assessments/projects/in-toto

lumjjb:
@ultrasaurus I believe the presentation to the TOC is over? shall we update it and close it out :)

ultrasaurus:
linked video in the issue description

PR #202, opened 2019-06-13
assessments/projects: add in-toto self assessment

SantiagoTorres:
This relates to issue #166 

This adds a markdown version of the self assessment for in-toto.

cc @ultrasaurus @JustinCappos @justincormack @lumjjb


assessments/projects/in-toto/self-assesment.md
```diff
@@ -0,0 +1,783 @@
+﻿# in-toto security self-assessment 
+## with input from the SIG-Security WG
```
ultrasaurus: suggest dropping this second-line since it is implied by contributors list below

assessments/projects/in-toto/self-assesment.md
```diff
@@ -0,0 +1,783 @@
+﻿# in-toto security self-assessment 
+## with input from the SIG-Security WG
+
+June 5, 2019
+
+*_Authors_*
```
ultrasaurus: line breaks below don't appear in rendered markdown, so I think it would be a little more readable if it was 

*_Authors_*: name, name, name 

keeping markdown source as separate lines for easier reading/diffing of source

assessments/projects/in-toto/self-assesment.md
```diff
@@ -0,0 +1,783 @@
...
+
+#### Internal
+We have a semi-private (publicly-writable) mailing list as a google group at
+in-toto-dev@googlegroups.com
+
+#### Inbound
+
+It is possible to reach out the in-toto team members at:
+
+* #in-toto@irc.freenode.net
```
ultrasaurus: is it possible to make this a link?  (otherwise in `...` so it doesn't appear as mailto, since I think that's incorrect)

assessments/projects/in-toto/self-assesment.md
```diff
@@ -0,0 +1,783 @@
...
+* The #in-toto channel at secure-systems-lab.slack.com
```
ultrasaurus: would be nice to link "secure-systems-lab.slack.com" to wherever it's listed how to get an invite.  If I just go there in slack, it says:
image

ultrasaurus: unless that is intended for core team only and public should access via in-toto freenode, if that's the case, just note that please

SantiagoTorres:
Thanks for the review @ultrasaurus. I think I addressed all the comments and I'm ready for another review :+1:

PR #204, opened 2019-06-13
starter proposal template

ultrasaurus:
to address https://github.com/cncf/sig-security/issues/197


JustinCappos:
I like the sparseness.  We don't want a lot of process around this.

PR #227, opened 2019-06-27
Revert "document currently active roles"

ultrasaurus:
Reverts cncf/sig-security#219

want to address comments before merge


ultrasaurus:
FYI -- I force-removed this from master, since it did not create the intended effect of allowing for addressing comments in the PR and re-merging

PR #236, opened 2019-07-10
Clarify and rename identity category

zollman:
Per #110 , proposing initial changes to clarify the Identity category, and rename it to Identity Lifecycle and Access Management. 

Hopefully this makes it clear that it's not just about each service's own entitlements or authn/authz libraries, but really focused on tools which make it easier for all workloads to support authentication, secure use of tokens, manageable entitlements and control the lifecycle and auditability of those.


landscape/categories.md
```diff
@@ -23,24 +23,29 @@ sub-categories. The remaining text is a description of each sub-category.
   - _Secret Management_
     - Tools that help to manage, encrypt or distribute secrets and other sensitive information.
 
-- **Identity & Access Control**
-  - _Identity_
-    - SPIFFE, identity providers, OpenID, LDAP, Okta
+- **Identity Lifecycle and & Access Management**
+  - _Identity Lifecycle_
+    - Systems for creating, governing and terminating identities and entitlements for users and systems
+      - Lifecycle management tools like Okta, FreeIPA
```
zollman: Do we want to name specific things here? Is this a governing document describing the category, or the root of discussion?

landscape/categories.md
```diff
@@ -23,24 +23,29 @@ sub-categories. The remaining text is a description of each sub-category.
...
+      - Service identity for identifying ephemeral workloads, e.g. Spiffe
```
zollman: If we wanted to include commercial entities, I'd point to commercial implementations like Scytale or Tremolo here as well.

landscape/categories.md
```diff
@@ -23,24 +23,29 @@ sub-categories. The remaining text is a description of each sub-category.
...
     - Controls within the orchestration layer to provide minimum required access to orchestrator resources.
-    - Authentication / Authorization
+    - Resource authorization
       - OAuth, IAM
-    - Privilege & Access Management
-      - Managing privileges in the platform for humans and machines, access controls, audit logging, privilege revocation,
-        etc
-    - Network Policy & Protection
-      - Firewalls, network segmentation, network policy
+      - Policy engines
```
zollman: OpenPolicyAgent belongs somewhere; I put it here.

landscape/categories.md
```diff
@@ -23,24 +23,29 @@ sub-categories. The remaining text is a description of each sub-category.
...
+    - Privileged Access Management
+      - Applying higher levels of control for sensitive actions, including just-in-time access and enhaned logging
   - _Service access_
     - Service admittance & admission controllers
       - Allowing new service instances to join existing services, and preventing insecure application instances from running.
     - Storage access
+      - Governing access to stored data by volume, schema, record or field
```
zollman: Data security at a lower level than volume - databases and object/file-level storage - are increasingly important for companies that want to run "data lakes" and there are a bunch of new commercial vendors for these systems. I combined "storage access" for all types of storage but eventually I think we'd want to parse it out.

landscape/categories.md
```diff
@@ -23,24 +23,29 @@ sub-categories. The remaining text is a description of each sub-category.
...
+    - Network Policy & Protection
+      - Firewalls, network segmentation, network policy
   - _Key and Certificate Management_
```
zollman: Considered renaming this to "authenticator management" using the NIST definition (https://csrc.nist.gov/glossary/term/authenticator) - tokens, passwords, keys and certificates - but I don't know how many people are familiar with that word. Keys and certificates is probably clearer.

lumjjb: The addition of network here is great.

landscape/categories.md
lumjjb: Taking from my comment in https://github.com/cncf/sig-security/issues/234, I'm not a fan of the naming of this being restricted to the particular activities, but I think putting it down like you've done here is good. Perhaps if we presented it as identity related activities including x, y, z.

landscape/categories.md
lumjjb: I don't think we want to have too many details of specific projects here since the document is more of a description of how to classify things... And based on that, we would come out with a landscape document that actually contains the actual implementations/projects of the ecosystem.

lumjjb:
I really like the details that you've added to the different categories of the document. 

I think there are discussion points that would be helpful to get more feedback on - about whether we want to include names of specific projects (and to what extent), as well as the identity definition. We could put this on the agenda the next call.

zollman:
Newly clarified after feedback from sig-security QWG

landscape/categories.md
```diff
@@ -24,20 +24,34 @@ sub-categories. The remaining text is a description of each sub-category.
...
+  - _Access control and enforcement engines_
+    - Attribute and policy-based resource authorization engines and libraries
+      - Including general policy engines
+    - Role-based access control engines
+      - Including libraries and standards for common roles
     - Privileged Access Management
-      - Applying higher levels of control for sensitive actions, including just-in-time access and enhaned logging
-  - _Service access_
+      - Applying higher levels of control for sensitive actions, including just-in-time access and enhanced logging
+  - _Resource-specific access layers_
```
lumjjb: Could you clarify what you mean by layer? Layer implies some sort of stack.

lumjjb: @zollman could you clarify on this and then we can probably merge this in :)

izgeri:
@lumjjb if you end up creating another PR with your suggested changes, please ping me for review.

lumjjb:
Closing in favor of https://github.com/cncf/sig-security/pull/274 due to inactive author.

PR #237, opened 2019-07-15
[WIP] Transcripts from presentations

ultrasaurus:
I haven't reviewed the docs, just submitting the updated transcript

I wonder if these should be in directory `case-studies` to go with https://github.com/cncf/sig-security/pull/195


lumjjb:
>I wonder if these should be in directory case-studies to go with #195

maybe we could have soft links from the case directory directory to the transcripts directory? 

The transcripts look good. How did you do this? is it a zoom service?

lumjjb:
@ultrasaurus do we want to merge this? Or only when we have a reliable and consistent way to do transcript uploads?

PR #242, opened 2019-07-25
Consolidated Discussion on Policy Formal verification

rficcaglia:
This is a write up of a discussion that started a couple months ago regarding formal verification.  This narrows the scope to policy verification, akin to SecGuru, AWS Zelkova and Tiros.

Discussed on the Policy WG call 7/24/19.


rficcaglia:
When merged, closes https://github.com/cncf/sig-security/issues/196

cloudsriseup:
Looks good to me.

ultrasaurus:
It's good to see work from the policy sub-group make its way into the repo!  I just had time to skim the doc and added a quick note in-line -- a few nits below on repo style...

* I don't think I saw a PR that addresses the definition of "policy-wg" -- I think there was an open issue to write that up (cc @hannibalhuang ) -- maybe just make the directory /policy for now and we can keep policy-related docs here

* pls reformat with line breaks after 80 chars -- makes future diffs more readable, see https://github.com/cncf/sig-security/blob/master/CONTRIBUTING.md#writing-style

Thanks!

lumjjb:
The document is written as a first person, so probably you may want to add your name as the author in the document as well (unless i missed it somewhere).

rficcaglia:
@lumjjb @ultrasaurus I have addressed your suggestions. apologies for _also_ adding additional material in the same commit, but hopefully this updated content is acceptable and can be merged in.

@hannibalhuang @ericavonb will review on today's policy call (I think we have one scheduled, yes?)

scotty-c:
@rficcaglia This is a really good write up and a great starting point.

PR #243, opened 2019-07-26
Adding the logo files from cncf artwork repo

amye:


caniszczyk:
@amye just reference the logo from https://github.com/cncf/artwork - no need to add the binaries here

amye:
#thorough

Issue #165, opened 2019-05-16
Investigate the "platform implementer" persona


ckemper67:
Large cloud users frequently have a "platform team" that is responsible for ensuring that the organization's security and compliance constraints can be met. This is increasingly difficult in a hybrid world where the platform team now needs to track and understand the security and compliance offerings from multiple cloud providers. These "platform teams" need to make sure that guidance and sometimes services are in place that allow the administrator personas to be shielded from the differences across cloud providers.

Cloud Native technologies can provide a common ground to some extent, but they do not shield the administrators from being exposed to the platform specific implementations if the deployment consumes other services provided by the platforms. Blobstore is a frequently cited example and we have CNCF efforts like Rook that can provide an abstraction.

lumjjb:
Creating a link to https://github.com/cncf/sig-security/issues/109.

rficcaglia:
in terms of validating policy "overlap" and maintaining a global overall policy while allowing delegated policy....that was a use case we briefly discussed in the context of https://github.com/cncf/sig-security/issues/196 on the Policy-WG call.

lumjjb:
Resolved by https://github.com/cncf/sig-security/pull/246

PR #246, opened 2019-07-31
Add the use cases for the platform implementer persona

ckemper67:
To address #165, this PR adds the use cases for the platform implementer persona. The use cases are interesting, because the platform implementer's responsibility is to make some of the other roles successful.

To highlight that relationship between the roles, I decided to bold the use of other personas in the use case description.


usecases.md
```diff
@@ -160,8 +161,18 @@ impact of repeated request on the rest of the infrastructure.
...
+* As a platform implementer I can make my organization's **security administrators and compliance officers** successful.
+* As a platform implementer I can provide central administration of cloud resources to my **operators and administrators**.
+* As a platform implementer I can provide compartmentalization of cloud resources for delegation purposes by **operators and administrators**. 
+* As a platform implementer I can allow my **operators and administrators** to delegate administration of resources.
+* As a platform implementer I can enable my **security administrators** to constrain the behavior of **developers** by setting guardrails.
+* As a platform implementer I can enable my **security administrators and compliance officers** to enforce auditing of access policies. 
+* As a platform implementer I can enable my **security administrators and compliance officers** the auditing of resource access.
+* As a platform implementer I can enable my **compliance officers** the setting of audit logging policies for my organization's resources
+* As a platform implementer I can enable my **compliance officers** to certify the non-violation of my organization's compliance needs.
```
lumjjb: nit: spacing between bullets for consistency

usecases.md
```diff
@@ -160,8 +161,18 @@ impact of repeated request on the rest of the infrastructure.
 
 * As an end user, I can understand the effect of changes to policy that I am making
 
-
-
+# Platform Implementer
+
+* As a platform implementer I can make my organization's **enterprise, network and quota operators** successful.
+* As a platform implementer I can make my organization's **security administrators and compliance officers** successful.
```
lumjjb: These two points seem to be a summary of the below points. Organizing it into the two groups could be very helpful for readability.

usecases.md
lumjjb: My assumption of a platform implementer is that he/she should be responsible across multi-cloud or hybrid cloud boundaries - like in the security administrator case. Or is the platform implementer here scoped to a single cloud?

ckemper67: No I think the multi-cloud and hybrid deployments are actually very relevant to this persona, maybe to all personas here. That might be something to put into the intro paragraph.

usecases.md
ckemper67: Do you mean one group each for operators and administrators or one group each for summary and details?

usecases.md
ckemper67: ack

ckemper67: I decided to add a dedicated intro paragraph for the platform implementer persona and specifically mention the bridging of heterogeneous platforms.

usecases.md
lumjjb: one group each for operators and administrators

lumjjb:
@pragashj any thoughts on the addition?

lumjjb:
Merging as discussed on 09/18 call.

PR #253, opened 2019-08-13
Adding the SIG-Security color scheme

amye:
Adding SIG-security-color-scheme.pdf


ultrasaurus:
Would it be possible to format this in HTML or something a bit easier to view than PDF?  (Hard to copy text / color values from PDF)

lumjjb:
I was thinking of whether we should merge this first and make a new PR to convert this to HTML, but I think it would not be good to add tracking of a large blob like that. Should we create a suggestion to see if anyone wants to take up the task of converting this?

alexcontini:
hi! I created the pdf and I thought something like this might work - https://github.com/alexcontini/sig-security/tree/master/logo ? Let me know if it looks okay or any feedback and we could either add that to the cncf repo in the sig security folder

lumjjb:
@alexcontini That looks awesome!! Could you create a PR (referencing this one) and we can review and merge it! Thanks!

lumjjb:
Addressed by https://github.com/cncf/sig-security/pull/339

PR #259, opened 2019-08-28
assessment matrix updated after 8/28 call

rficcaglia:
* removed Observer/intern column due to confusion (and lack of volunteers!)
* moved Falco to primary spot
* KeyCloak deferred
* put me on as default for Falco lead pending some other volunteer (hint!) @JustinCappos @ultrasaurus @dshaw up to you to decide (I was passively involved in the OPA assessment but recused myself from the sig-security side due to conflicts)


JustinCappos:
Thanks for the PR.  I think Falco will be interesting to assess.  I'm torn between participating and just looking at the results.  

@rficcaglia leading the assessment sounds great to me.  Do others have thoughts?

MVrachev:
> * removed Observer/intern column due to confusion (and lack of volunteers!)

I don't agree that this is the solution.
The unclarities surrounding that `Observer/Intern` role isn't a good argument to the abandonment of this role.

The requirement for a `Security reviewer` role is:
`Unless approved by SIG-Security chairs, at least one of the reviewers will have previously performed a CNCF security assessment.` 
How can you achieve that without giving members the chance to be an `Observer/Intern` on a few assessments and then to become a Security reviewer?

If we want the security assessments to become a more regular thing in future then there should be an `Observer/Intern`.

About the lack of volunteers: is there a security assessment when the `Observer/Intern` role was advertised? I haven't heard of any. 
Also, the unclarities surrounding this role are another reason why there isn't a big interest for now.

I am a volunteer for being an `Observer\Intern` in one of the next assessments.

rficcaglia:
Huge thanks for volunteering!

My intent in the PR was not to delete the role, but minimize confusion on
the grid, ie everyone is a “volunteer” but some volunteers are interns,
some volunteers are experienced. For overview/archival purposes it probably
doesn’t Matterhorn we present it in the grid. However, noting the lead
separately makes sense as a point of contact.

That said happy to add back the intern column if it was helpful.

Re: increasing the number of volunteers. I think once Kris confirms timing
for Falco, yes we can send out a blast or something to advertise for more
volunteers.  Also with OPA’s assessment nearly complete, we could use that
to generate some interest...eg add a pitch for volunteers at the end of all
TOC or other SIG presentations.

lumjjb:
I think to avoid confusion, we should have this matrix reflect the terminology used in the guide, so instead of `Volunteers`, it would be `Security Reviewer` and `Lead Security Reviewer`. 

I think we have a good discussion in https://github.com/cncf/sig-security/issues/256, and we can draft an initial guide based off that and refine it as we go along - like how we did for security reviewer definition :). When we get an initial draft of this in, we agree on advertising on a weekly sig meeting.

lumjjb:
>put me on as default for Falco lead pending some other volunteer (hint!) @JustinCappos @ultrasaurus @dshaw up to you to decide (I was passively involved in the OPA assessment but recused myself from the sig-security side due to conflicts)

Sorry was out last week. From what I recall, initially it was @justincormack lead for Falco and I was lead for Keycloak. But because of RH acquisition, it created a conflict of interest. So Justin and I swapped lead roles for the project. It's not cast in stone, so i'm good to revisit this discussion since time has passed since we last discussed it.

rficcaglia:
@lumjjb I'm happy to be in either role - if you want to take Lead let me know.

changed the col headers per your suggestion and updated reviewers.

JustinCappos:
> thanks @rficcaglia for updating this. Also appreciate the suggestion by @lumjjb about column
> names which I think is an improvement.
> 
> Added @JustinCappos as FYI. I'm going to merge this, since I think it is an improvement, then
> there should be follow-up PRs to adjust Falco reviewer list.

Sounds good!

PR #266, opened 2019-09-09
add Justin Cormack as a meeting facilitator

justincormack:
Signed-off-by: Justin Cormack <justin.cormack@docker.com>


ultrasaurus:
Post-note:  missed on pre-requisite about serving in a leadership role

@justincormack write the initial draft of the SIG-Security charter (to help TOC transition from WG to SIG structure).  This was not very visible, but important foundation work that I was also involved in, so can attest to his contribution.

Issue #309, opened 2019-12-14
Is the assessment matrix duplicative with the issue process for assessments?


JustinCappos:
It feels like the assessment matrix: https://github.com/cncf/sig-security/blob/8f02ab936dd1ec05405ca995221b601c5a7382ca/assessments/assessment-matrix.md
is largely duplicative with the tagged items in the issue tracker:  https://github.com/cncf/sig-security/labels/assessment

Do we need both?  If so, could we generate from the other (likely issue -> matrix)?

ultrasaurus:
see related discussion "Assessments Listing" (issue https://github.com/cncf/sig-security/issues/206)

I think the [current matrix](https://github.com/cncf/sig-security/blob/8f02ab936dd1ec05405ca995221b601c5a7382ca/assessments/assessment-matrix.md) could be simplified to be a priority queue, addressing the remaining open issue in the "intake process" PR (https://github.com/cncf/sig-security/pull/296)

ultrasaurus:
What about if we made a github "project" for assessments?  it would be easier to maintain the order of the queue (since it seems the start of an assessment is often delayed due to self-assessment not being ready)

JustinCappos:
> What about if we made a github "project" for assessments? it would be easier to maintain the order of the queue (since it seems the start of an assessment is often delayed due to self-assessment not being ready)

I'm not 100% sure what you mean.  Is the purpose of the project to give a template for self-assessment?  What is it for?

ultrasaurus:
I created a TEST project to illustrate: https://github.com/cncf/sig-security/projects/2

JustinCappos:
I created the need-self-assessment label to help us track.  Looks like we can close this!

ultrasaurus:
I think we can remove the [assessment matrix](https://github.com/cncf/sig-security/blob/8f02ab936dd1ec05405ca995221b601c5a7382ca/assessments/assessment-matrix.md) page then

lumjjb:
I love the new project board! :)

JustinCappos:
Great!  After #310 is approved and merged, we can close this issue then...

PR #310, opened 2019-12-16
Removing redundant assessment matrix #309

JustinCappos:


ultrasaurus:
@Robert Ficcaglia it would be great to have your review on this -- I think that we've now evolved the issue template and with new project board (https://github.com/cncf/sig-security/projects/2) that we have covered all of the content here

ultrasaurus:
https://github.com/cncf/sig-security/issues/309 fixed!

Issue #233, opened 2019-07-03
roles and responsibility for governance


ultrasaurus:
in roles.md...
"Members who take on Roles are responsible for
understanding the approval policies related to their role and ensure that
commits to the repo are reviewed in accordance with our [governance](./)"

via [feedback thread](https://github.com/cncf/sig-security/pull/229/files/477a6a607be4ec4b6103339ba9d570e88c0858db#diff-d2e08ed5f897b2157042d4c3bfc22eb9)

@TheFoxAtWork: not sure if "governance" is the same as the approval policies in line 118, if so we should be more clear "...the approval policies, as defined in our governance,

@ultrasaurus: well, the security reviewer role is defined outside of /governance -- just trying to say that anyone with a role that involves commits to the repo is responsible for anything that is written in our governance (e.g. consistent with charter and mission, transparency, inclusion, and anything else that we wrote down as overall policy for how we work together as a group)

Feel free to suggest alternate wording, but I don't want to imply less responsibility for people who take on roles. I want people to feel an obligation to fully understand and commit to taking ownership of understanding our overall governance and acting appropriately.

TheMoxieFox:
Perhaps something like " Members fulfilling any Roles in SIG-Security are responsible for understanding and abiding by the governance and policies defined by SIG-Security.  This commitment and execution of understanding includes not only commits to the repo, but also to any approvals or direction required by their Role."

ultrasaurus:
I like this wording.  Added 'triage' label for visibility.  

From my perspective, all that's needed on this issue now is a PR :)

TheFoxAtWork:
read for review and merge upon approval @ultrasaurus

lumjjb:
@TheFoxAtWork , @ultrasaurus Have all the items in this issue been resolved and we can close this issue?

TheFoxAtWork:
i think this can be closed

PR #316, opened 2019-12-19
Roles responsibility

TheFoxAtWork:
this is the PR for ticket #233


ultrasaurus:
this seems to have supply-chain change included... prolly not intentional?

TheFoxAtWork:
ugh!  doing too much too quickly so sry!

PR #339, opened 2020-02-11
add logo and design page

alexcontini:
#253 

Here is the PR to add logo and design page!

@lumjjb


alexcontini:
@lumjjb great! and sorry don't know where those came from. They should all be deleted now

lumjjb:
> @lumjjb great! and sorry don't know where those came from. They should all be deleted now

Thanks! Those are auto generated by Finder in mac. All good.

lumjjb:
@alexcontini I think i still see a couple more in some of the sub folders: like `logo/sig-security/icon/color/.DS_Store`

lumjjb:
Looks like 1 more and we should be good: `logo/sig-security/horizontal/.DS_Store`!

alexcontini:
@ultrasaurus  ah yes, that is an issue, confused as they line up on mine, so I need to look into this. I will have a fix for it very soon and will comment when so!

Here is what I am seeing, no idea why:


Screen Shot 2020-02-14 at 5 51 42 PM

lumjjb:
I think spacing isn't quite as consistent as one would like, especially with the interaction of different zoom levels. I think the zoom of the text and zoom of pictures are slightly different. I think one way to do this is to manually draw the boxes in and color them with CSS. 

I think we can merge this in first since it is a non-trivial fix (at least in my head :)). We can merge this first and open an issue to fix this.

Issue #343, opened 2020-02-24
where to put assessment summary slides in the repo?


ultrasaurus:
I don't think the [assessment summary slides](https://docs.google.com/presentation/d/1Cinwa2grYSdP4yNQKS1ICIBH3oSkBPgDZl2RlG5asRA/edit#slide=id.g5a0cdb412d_0_0) are linked from anywhere...

I would prefer an open format, but don't know of one that works well for slides.  In the meantime, these should be linked somewhere

ultrasaurus:
@JustinCappos @TheFoxAtWork @lumjjb @rficcaglia  thoughts about where these should be linked?

TheFoxAtWork:
@ultrasaurus I can see them either going in the individual project folder under assessments; [i.e assessments>projects>in-toto](https://github.com/cncf/sig-security/tree/master/assessments/projects/in-toto) , or under the projects folder to enumerate ones that are complete; [i.e. assessments>projects](https://github.com/cncf/sig-security/tree/master/assessments/projects)

rficcaglia:
no strong opinion on where - I myself would think to look in
assessments>projects>project first.

as to open format haven't tried pandoc but maybe we can convert
markdown<->slides: https://pandoc.org/

lumjjb:
Yea, i second the caution towards using binary blobs as they don't generally deal well with git diffing and history. I think having links initially to 1 set of slides that have all the 1 slide reviews would be okay to start with.

TheFoxAtWork:
Can we re-scope this to just including the link in assessments>projects>project per @rficcaglia 's comment?  we could add it under Step 5 of the [Assessments guide](https://github.com/cncf/sig-security/tree/master/assessments/guide), simple enough to get started.  that would close out this issue.  If we need to we can address standard formatting as a separate issue.

TheFoxAtWork:
This change is included in the #449 PR's (#479 & #488)  once those are merged we can close this.

anvega:
https://github.com/cncf/tag-security/pull/479 & https://github.com/cncf/tag-security/pull/488) merged

Issue #341, opened 2020-02-14
conflict of interest review unclear 


ultrasaurus:
The intent was to only require chair review if there were hard-conflicts or the security assessment facilitator flagged a concern of too many soft conflicts -- this is not at all clear in how we wrote it down and should be fixed.  

In the meantime, let's not block assessments on chair review if @JustinCappos agrees that there are no blockers and the project team is ready to move forward.

https://github.com/cncf/sig-security/issues/308#issuecomment-568119151

lumjjb:
I think when I wrote the text before, my initial thinking was the same as the comment, so perhaps we got lost in translation somewhere. I do agree that probably chairs should not be required to unblock if there are no complications. 

Perhaps we can specify responsibility of the security lead reviewer to also ensure all conflicts and specified before proceeding to review, and to raise to security assessment facilitator if there are any conflicts. And from there its either check off by @JustinCappos or if it judgement says someone should not be on the project then it gets raised up to chairs?

ultrasaurus:
working on this

TheFoxAtWork:
closing as this was merged

Issue #326, opened 2020-01-08
[Suggestion] Updates to Lead Security Reviewer Role and Security Assessments Guide and phrase naming


TheFoxAtWork:
Description: 

**1 - Clarifying Question changes**
- [ ] Change Dumb question phase to Clarifying Question Phase in assessment guide
- [ ] Update Clarifying Question Phase as the responsibility of the Lead Security Reviewer explicitly or their designee 
- [ ] explain what Clarifying Question phase is
- [ ] Clarifying question phase is adhoc and not in the 3 week cycle.

**2 - Lead Security Reviewer Changes**
- [ ] Update Lead Security Reviewer role to state Clarifying Question Phase is their responsibility 
- [ ] explain why limited to them or their delegate
- [ ] Clarifying Question phase is also an initial pass before other reviewers to ensure document content is clearly understood, terms explained, concepts have context.  This enables reviewers to focus on security and technical details.  Provide option to delegate this function.

**3 - Reviewer Role Changes**
- [ ] Update other Reviewer roles to exclude Clarifying Question Phase unless delegated
- [ ] point back to what the phase is (clarifying question) and who is responsible.

**4 - Assessment Guide Changes**
- [ ] Update assessment guide with more content: suggestion to review the repo
- [ ] list where final assessment slides are placed in the repo as part of the security assessment guide/outline/expectations.
- [ ] clearly define format for conflict of interest
- [ ] explain importance of conflict of interest statements
- [ ] identify [conflict review phase](https://github.com/cncf/sig-security/issues/341)

**5 - Self-Assessment Template Updates** (outline)
- [ ] Update the self-assessment template to encourage cross-linking to relevant sections of the repo or other docs
- [ ] update self assessment with section for threat model to ensure projects have the opportunity to include it.
- [ ]  include section for Critical Functions & Features that are security relevant to the software.   These items could then be leverage for future reviews under change detection as well as manageable items for a security audit.

TheFoxAtWork:
also need to add more content to the security assessment guide, reviewing the repo.  Also maybe update the self-assessment template and include language to encourage cross-linking to relevant sections from the repo.

TheFoxAtWork:
Updates to include the lead reviewer doing an initial pass to ensure all other reviewers can clearly understand the content of the document, what do the terms mean?  what is the context for the concepts introduced, etc.  this allows reviewers to ask more pointed technical questions of the assessment

TheFoxAtWork:
[...]

Issue #171, opened 2019-05-22
Threat modeling 


tpepper:
KubeCon EU presentation on security audits mentioned a finding / recommendation for projects to understand their threat model, but this feels overly abstract.

Should an input criteria to a CNCF sec audit be that the project have a threat model documented? CNCF is suggesting projects follow the LF’s CII best practices recommendations, but that doesn’t require a threat model.

Even if not required, I suspect threat modeling education by the SIG out to it’s associated projects would have value as many many software engineers are unfamiliar with what establishing one entails and exceptionally few projects have them.

omerlh:
I think this is a really good proposal, we should at least recommend the project to have it - and maybe set it graduation criteria. There are some discussions on how to document, I have some example I wrote [here](https://kamus.soluto.io/docs/threatmodeling/threats_controls/) and will be happy to hear feedback about the format and readability.

lumjjb:
I believe we do ask for some threat modeling when performing security reviews for the CNCF projects - is that along the lines of what you have in mind?

omerlh:
Interesting, where does it documented?

JustinCappos:
This is listed in the security assessment docs that the project submits to the SIG-Security working group.  I'll list a few examples here:  
https://docs.google.com/document/d/1O3O7eFCTZ_7lLZ9IhHSWFiXhjWf2gPuJbzZ7u0m-9uI/edit?pli=1#heading=h.oykub2ep8ndw
https://docs.google.com/document/d/1L15ZZXnDxe3eqXnKHiZRxdr59tGkt_ipPixdA6RbzII/edit#heading=h.4zw236cia0t9

TheFoxAtWork:
This is issue is completed initially with PR #349.

PR #349, opened 2020-02-27
Issue 326 assessment updates

TheFoxAtWork:
updated with #341 , #343, #326 , #171


lumjjb:
Branch from master instead of spiffe branch

Issue #353, opened 2020-03-05
[Suggestion] Documentation update - roadmap.md


IAXES:
Description: The "SAFE WG Roadmap" title should be updated to "SIG-Security Roadmap" [in the roadmap document](https://github.com/cncf/sig-security/blob/master/roadmap.md), as per [a related discussion in Slack](https://cloud-native.slack.com/archives/CDJ7MLT8S/p1583401459049600).

Impact: Minor documentation updates.

Scope: Minimal.

TheFoxAtWork:
updated with new PR

PR #354, opened 2020-03-05
Issue 353 safeeg

TheFoxAtWork:
#353


lumjjb:
I think you branched from the spiffe branch. Can you branch from master instead?

lumjjb:
I think you can just rebase master instead, that should work

PR #356, opened 2020-03-05
TechLead: link to CNCF process

ultrasaurus:
there should be more detail here, but not sure that the ad hoc process that we
followed this time should be *the* process, so leaving intentionally vague for now


ultrasaurus:
governance changes need 2/3 chair review -- the actual process change happened when we joined CNCF as official SIG, but still want JJ and/or Dan to review (I think my submission of the PR counts as a review)

cc new tech leads @lumjjb @TheFoxAtWork @JustinCappos -- FYI we wanted to keep process light for now while we sort out more details about how we coordinate responsibilities across Tech Leads and chairs.  Eventually we want to either expand the number of tech leads or cycle in new folks so that we have more geographic diversity (across timezones and KubeCon regions).

lumjjb:
Hmm. thoughts on linking to this page? https://github.com/cncf/toc/blob/master/sigs/cncf-sigs.md

Which has the generic CNCF process.

ultrasaurus:
FYI -- I linked to CNCF README, since I that links to page that @lumjjb refers to and thought it would be more likely to be the canonical link.

cc @lizrice @justincormack

governance/roles.md
```diff
@@ -65,11 +65,8 @@ The general list of activities for TL are:
   * Propose agenda items for meetings to ensure that open issues are
   discussed with the group when needed
 
-TLs are elected by the following process:
-  1. One of the Chairs nominates a candidate to be a TL, and creates a PR that
-  adds the TL to the [SIG README](/README.md).
-  1. Members are invited to comment and vote according to the regular voting
-  process described below.
+TLs are assigned by CNCF Technical Oversight Committee
```
dshaw: Per the CNCF Special Interest Groups ("SIGs") charter, the TOC and Chairs nominate Tech leads.

governance/roles.md
```diff
@@ -65,11 +65,8 @@ The general list of activities for TL are:
...
+(see [CNCF TOC SIGs](https://github.com/cncf/toc#sigs)).
```
dshaw: (see [CNCF SIG Tech Lead nomination and election process](https://github.com/cncf/toc/blob/master/sigs/cncf-sigs.md#elections)).

governance/roles.md
ultrasaurus: The key difference is that tech leads require a vote of the TOC (not the SIG membership or chair majority.  Maybe simply a reference without an attempt to summarize, because the TOC process isn't super clear about what happens before the vote (and I think there's a PR open on that)

ultrasaurus:
This needs editing to address Dan's feedback and my initial edit -- neither are quite right.  Would appreciate alternate suggested edit.

lumjjb:
I feel like just linking the TOC process is sufficient. It seems like the process is governed by TOC, unless we have a good reason to add exceptions to it.

Issue #348, opened 2020-02-26
Cloud Native Security Map (Landscape v2)


lumjjb:
Description: 

Provide a different view of a security landscape which will provide more useful information 

Impact: 

The current security landscape is a list of categories which are useful for an overview. We want to provide an alternative view of the landscape which can be consumable by security practitioners, executives, and users of security.

Scope: 

Create a visual landscape of multiple processes, "Creating a cloud native app", "Setting up a cloud native infrastructure", "monitoring, logging, remediation and alerting in cloud native". Each process would be decomposed to several steps and sub-processes. Each of which will detail information about relevant threats, examples and mitigations.

EDIT: from whitepaper discussions: The landscape should mirror the topics of the whitepaper, and provide a visual guide and act as a tool for practitioners to find the correct tools to help them.

TO DO
- [x] TAG Representative @lumjjb
- [X] Project leader(s)
 - Ash (@ashutosh-narkar)
 - Brandon Lum (@lumjjb)
 - Diego (@dcomas)
- [ ] TBD

lumjjb:
Here is a mockup that we've worked on, I'd like to be able to put this in an interactive web format.

https://drive.google.com/file/d/1sIBxUT1ohmQCzR_ms4OpJ_GAh96IEjZX/view?usp=sharing

If anyone has some expertise doing web work, would be great to get additional eyes on this!

yeelinglam:
@lumjjb - I could give this a try, please provide a bit more around how "interactive" you would like.  Also, do you want it to be hosted on Github or is there a target web hosting place?

lumjjb:
Awesome - I think we can work with the assumption that we can have other hosting. We can ask for some resources from the CNCF in that regard. 

I found this example that I really liked as part of gojs library (although I think we can't use it because its a commercial library). https://gojs.net/latest/intro/events.html (look at the first example of the page, where each word is clickable). 

Having the diagram with the flow on the top and whenever you click on one of the items, the description would just appear at the bottom of the diagram. The content would look something like this (ignore the words in the diagram at the top - those should just be each individual step) https://sig-security-scenarios.netlify.app/commits/

Clicking on one of the boxes in the flow would lead to the bottom updating with the details of that box. It would be added bonus to having nice navigation of the diagram.

I may not be very good with articulating what I'm thinking graphically, so if you'd like we can jump on a call to chat.

yeelinglam:
[...]

PR #383, opened 2020-05-15
Update cncf-projects.md - project list link(s)

whaber:
Should the project list contain links to each project?  (I updated the first one as an example).

If this is merged, I will make updates to the others.

### DIFF
--- a/governance/cncf-projects.md
+++ b/governance/cncf-projects.md
@@ -9,7 +9,7 @@ each project's annual [security assessment](/assessments).
 
 Current list of projects:
 
-* Falco
+* [Falco](https://falco.org/)
 * Open Policy Agent
 * Notary
 * TUF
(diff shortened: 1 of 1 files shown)


lumjjb:
This looks fine! Thanks @whaber , go ahead and do it for the entire list!

whaber:
@lumjjb I added descriptions because each project's purpose is not well known.  It helps to gain interest in those projects.  That being said, the descriptions will change over time which will cause this content to lag.  I think the pros outweight the cons.  However, if you disagree, I can remove the descriptions.  Thanks.

lumjjb:
What you've said is valid, and is actually part of what the security landscape hopes to do. (https://github.com/cncf/sig-security/blob/master/landscape/approach.md) - We are currently working on a revised Landscape v2, if you're interested to work on this (https://github.com/cncf/sig-security/issues/348)

I think the intent of this list is misunderstood. This list is not meant to be a resource for a link to a list of security projects. This is a list which the TOC specifically has made asks for us - which is used to determine assessment priority. This list is not meant to be a comprehensive list of security projects.

whaber:
Thanks @lumjjb  I will close this and take a look at #348

PR #390, opened 2020-06-03
extend dshaw term

ultrasaurus:
due to COVID we are increasing chair term +3 months -- thank you @dshaw for your continued service!


ultrasaurus:
@dshaw graciously agreed to extend his term while we figure out who will be our next chair
discussed with co-chairs @pragashj  and TOC liaisons: @justincormack @lizrice 
cc Tech Leads: @TheFoxAtWork @JustinCappos @lumjjb

Issue #399, opened 2020-06-12
[Suggestion] assessment intake process update


TheFoxAtWork:
Description: Update assessment/intake-process to account for pro-active projects looking to receive assessments from SIG-Security that are not TOC directed.

Impact: Provide clear expectations in order to assist in assessment queue management and project engagement

Scope: Minimal, discussed previously with co-chairs and tech leads. Just need to update the docs with what we've been doing.

TheFoxAtWork:
closing as PR was merged.

PR #400, opened 2020-06-12
updated per #399, need @justincappos should definitely review

TheFoxAtWork:
updated intake process per last tech lead/co-chair regarding CNCF/non-CNCF projects requesting review.  gives the area lead authority to kick them from the queue if the are lead determines they are no-longer communicating.

### DIFF
--- a/assessments/intake-process.md
+++ b/assessments/intake-process.md
@@ -57,8 +57,9 @@ charter.  In all cases, the priority queue will be maintained transparently as
 described below, along with communication via regular chair-liaison meetings and
 SIG-Security reports at TOC meetings.
 
-The Security Assessment Facilitator with concurrence from one (1) co-chair will
-determine if a CNCF or non-CNCF project may be accepted into the assessment
+A project may be accepted into the assessment queue, either by the Security Assessment 
+Facilitator with concurrence from one (1) co-chair, or by two (2) co-chairs.  This concurrence 
+is given by commenting on an issue proposing that the project be added to the assessment
 queue.  If at any time, the project's requesting review cease communicating, the
 Security Assessment Facilitator may remove the project from the queue with
 notification to the co-chairs.  The Security Assessment Facilitator will update
--- a/assessments/intake-process.md
+++ b/assessments/intake-process.md
@@ -10,10 +10,10 @@ assessments.
 
 # Authority
 
-Team members are welcome to submit PRs to streamline process when priorities
-are clear based on the criteria below. As needed, specific leaders (identified
-in the [repo root README](/README.md#security-assessments)) will coordinate
-the decision-making process.
+Team members are welcome to submit PRs to streamline this process when
+priorities are clear based on the criteria below. As needed, specific leaders
+(identified in the [repo root README](/README.md#security-assessments)) will
+coordinate the decision-making process.
 
 * The [Security Assessment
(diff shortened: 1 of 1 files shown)


assessments/intake-process.md
```diff
@@ -50,12 +50,19 @@ assessment, exceeding the bandwidth of the group:
...
-chair-liaison meetings and SIG-Security reports at TOC meetings.
+The Security Assessment Facilitator, in collaboration with the named chair, has
+the discretion to adjust priority in order to streamline the process, or per
+their own judgement for other reasons consistent with SIG-Security mission and
+charter.  In all cases, the priority queue will be maintained transparently as
+described below, along with communication via regular chair-liaison meetings and
+SIG-Security reports at TOC meetings.
+
+The Security Assessment Facilitator with concurrence from one (1) co-chair will
+determine if a CNCF or non-CNCF project may be accepted into the assessment
```
JustinCappos, line 61:
```suggestion
A project may be accepted into the assessment queue, either by the Security Assessment 
Facilitator with concurrence from one (1) co-chair, or by two (2) co-chairs.  This concurrence 
is given by commenting on an issue proposing that the project be added to the assessment
```

assessments/intake-process.md
```diff
@@ -50,12 +50,19 @@ assessment, exceeding the bandwidth of the group:
...
+queue.  If at any time, the project's requesting review cease communicating, the
+Security Assessment Facilitator may remove the project from the queue with
```
JustinCappos, line 63: I guess that marking the issue is a way to do this?  I just tried the process on Falco, but really no one got a chance to weigh in.  I don't know if this is what we want or not.

TheFoxAtWork: I think we want to give you enough leeway that if you find it is stalled with no responses to inquiries, you should be empowered to manage the queue accordingly, which may mean kicking them from the queue with public posting as to why ( a la issue update).  Definitely marking the issue.  
If you are looking for something more process specific with timelines how does the following look?
* If after 3 communication attempts by the Security Assessment Facilitator, Lead Reviewer, or a co-chair and the project ceases to respond the Security Assessment Facilitator may comment on the assessment issue and if no further updates are posted within 7 days, they will close the issue with notification to the co-chairs.

JustinCappos: I'd rather not be so prescriptive about the communication / timeline.  (If we were going to automate it, this is fine, but I won't remember these exact rules.)

I was more trying to point out that there wasn't really a chance for Falco to respond.  That may be fine, but we should be aware...

assessments/intake-process.md
```diff
@@ -50,12 +50,20 @@ assessment, exceeding the bandwidth of the group:
...
+the discretion to adjust priority in order to streamline the process, or per
+their own judgement for other reasons consistent with SIG-Security mission and
+charter.  In all cases, the priority queue will be maintained transparently as
+described below, along with communication via regular chair-liaison meetings and
+SIG-Security reports at TOC meetings.
+
+A project may be accepted into the assessment queue, either by the Security Assessment 
+Facilitator with concurrence from one (1) co-chair, or by two (2) co-chairs.  This concurrence 
+is given by commenting on an issue proposing that the project be added to the assessment
+queue.  If at any time, the project's requesting review cease communicating, the
```
JustinCappos, line 63:
```suggestion
queue.  If at any time, the project requesting review ceases communicating, the
```

Issue #407, opened 2020-07-22
[Proposal] Updates to Supply Chain definitions


TheFoxAtWork:
Description: Need to add attack chaining to Supply chain and update negligence definition to include "ack of adherence to best practices"
Impact: many supply chain attacks are a perfect storm of events that allow attacks to occur, they encompass multiple exploits across several areas.

To Do: create new listing in [definitions](https://github.com/cncf/sig-security/blob/master/supply-chain-security/compromises/compromise-definitions.md)

From PR [324](https://github.com/cncf/sig-security/pull/324) comments

knowlengr:
It's arguable, but a pretty thorough coverage of supply chain dependency appears in [NIST 8179,](https://nvlpubs.nist.gov/nistpubs/ir/2018/NIST.IR.8179.pdf) and that can be connected to the [NIST Risk Management framework](https://www.nist.gov/cyberframework/risk-management-framework). If there is a separate cloud native angle here, we may need to call it out.  The dependency model concern, though, is important, and if this is the only -- if inelegant -- way to surface it, then go for it.

lumjjb:
@TheFoxAtWork does #408 resolve the issue that you mentioned? There is a mention about not following best practice/procedure (around where it used to just be typosquatting)

TheFoxAtWork:
@lumjjb yep this can be closed.

PR #408, opened 2020-07-23
Minor edits for the supply chain compromise catalogue

joshuagl:
Most notably this expands the definition of negligence and adds a definition of attack chaining, both suggested in #407. I've also added some notes on mitigating dev tooling attacks.

### DIFF
--- a/supply-chain-security/compromises/compromise-definitions.md
+++ b/supply-chain-security/compromises/compromise-definitions.md
@@ -10,7 +10,8 @@ Index:
 * [Publishing Infrastructure](#publishing-infrastructure)
 * [Source Code](#source-code)
 * [Trust and Signing](#trust-and-signing)
-* [Attack Chaining](#attack-chaining)
+* [Attack Chaining](#technique-attack-chaining)
+
 
 ## Dev Tooling
 Occurs when the development machine, SDK, toolchains, or build kit have been exploited.  These exploits often result in the introduction of a backdoor by an attacker to own the development environment.
@@ -26,6 +27,7 @@ Occurs due to a lack of adherence to best practices. TypoSquatting attacks are a
 
 _Reference(s)_:
 - [Slashdot Article Regarding PyPI Attacks](https://developers.slashdot.org/story/17/09/16/2030229/pythons-official-repository-included-10-malicious-typo-squatting-modules)
+
 
 ## Publishing Infrastructure
 Occurs when the integrity or availability of shipment, publishing, or distribution mechanisms and infrastructure are affected.  This can result from a number of attacks that permit access to the infrastructure.
@@ -52,5 +54,6 @@ _Reference(s)_:
 - [Mitre: Supply Chain Compromise](https://attack.mitre.org/techniques/T1195/)
 - [Wikipedia Code Signing](https://en.wikipedia.org/wiki/Code_signing)
 
-## Attack Chaining
-Many supply chain attacks are a perfect storm of events that allow attacks to occur. They encompass multiple exploits across several areas, which are chained together to enable the final attack. The attack chain may include, or conclude with, other supply chain attacks defined here, but often start with social engineering or a lack of adherence to best practices around securing publically accessible infrastructure components.
+
+## Technique: Attack Chaining
+Sometimes a breach may be attributed to multiple lapses, with several compromises chained together to enable the attack. The attack chain may include types of supply chain attacks as defined here. However, catalogued attack chains often include other types of compromise, such as social engineering or a lack of adherence to best practices for securing publicly accessible infrastructure components.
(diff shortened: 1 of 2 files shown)


supply-chain-security/compromises/compromise-definitions.md
```diff
@@ -48,4 +52,5 @@ _Reference(s)_:
 - [Mitre: Supply Chain Compromise](https://attack.mitre.org/techniques/T1195/)
 - [Wikipedia Code Signing](https://en.wikipedia.org/wiki/Code_signing)
 
-
+## Attack Chaining
+Many supply chain attacks are a perfect storm of events that allow attacks to occur. They encompass multiple exploits across several areas, which are chained together to enable the final attack. The attack chain may include, or conclude with, other supply chain attacks defined here, but often start with social engineering or a lack of adherence to best practices around securing publically accessible infrastructure components.
```
lumjjb, line 56: This seems more of a technique rather than a type of compromise. It doesn't quite "typecheck" with the above headers. I think this is valuable information, how about putting it as an addon to the paragraph of types of supply chains on top - kind of like here are some types of supply chain compromises, sometimes, a breach may be attributed to multiple lapses, etc.

supply-chain-security/compromises/compromise-definitions.md
```diff
@@ -10,22 +10,26 @@ Index:
...
 
+- _Mitigation_ - use of trusted binary repositories; verification of provenance (i.e. signatures) and/or integrity (i.e. checksums) of dev tooling downloads; bootstrapping dev toolchain from a minimal, trusted and auditable seed (ideally source code).
+
 _Reference(s)_:
 - [Mitre: Supply Chain Compromise](https://attack.mitre.org/techniques/T1195/)
 
 
 ##  Negligence
-Associated with TypoSquatting attacks, the developer failed to verify the requested package or source code was spelled correctly.  
+Associated with TypoSquatting attacks. Occurs due to a lack of adherence to best practices, when a developer failed to verify the requested dependency name was correct (spelling, name components, glyphs in use, etc).
```
lumjjb, line 25: I think it would be nice to have "Occurs due to a lack of adherence to best practices" to be at the start of it - to describe what negligence refers to in this context, and typosquatting  attacks are an example of such negligence.

supply-chain-security/compromises/compromise-definitions.md
```diff
@@ -10,22 +10,26 @@ Index:
 * [Publishing Infrastructure](#publishing-infrastructure)
 * [Source Code](#source-code)
 * [Trust and Signing](#trust-and-signing)
+* [Attack Chaining](#attack-chaining)
 
 ## Dev Tooling
 Occurs when the development machine, SDK, toolchains, or build kit have been exploited.  These exploits often result in the introduction of a backdoor by an attacker to own the development environment.
 
+- _Mitigation_ - use of trusted binary repositories; verification of provenance (i.e. signatures) and/or integrity (i.e. checksums) of dev tooling downloads; bootstrapping dev toolchain from a minimal, trusted and auditable seed (ideally source code).
```
lumjjb, line 18: Caps Use

supply-chain-security/compromises/compromise-definitions.md
TheFoxAtWork: @lumjjb It may be beneficial to list it here, even though this is technique rather than true compromise type, multiple compromises occurring together for the purposes of this catalog showcases multiple weaknesses at a top level.  Also, some of the attacks in the catalog are unclear in the reporting which compromise was the initial, or true start of the compromises.  
Perhaps not including it in a compromise, but as an annotation below the definitions and listing in the compromise type table at the top level to improve readability?

lumjjb: Yea agreed, suggested top, but bottom is fine too, has the same effect.

supply-chain-security/compromises/compromise-definitions.md
joshuagl: Thanks for the suggestion. I've rephrased this a bit, please let me know what you think.

supply-chain-security/compromises/compromise-definitions.md
joshuagl: Thank you both for the feedback. I'm not quite sure what to do here. So far as I know, we can only anchor on headers so in order to include attack chaining as a clickable link in the compromise type list we need a header to link to.

We could add a heading level 1 above attack chaining? Or change the attack chaining header to include technique in the title somewhere? I've pushed a commit to this PR which follows the second option, adding "technique" to the header, and cleans up the description a bit.

joshuagl:
I've pushed some additional patches to try and address feedback here. I'm not sure if I've completely addressed the comment on technique vs. type on attack chaining. I'd welcome additional feedback on that.

TheFoxAtWork:
@joshuagl this looks great!!  @lumjjb can you take a look?

SantiagoTorres:
LGTM! I was going to propose the second option regarding [this](https://github.com/cncf/sig-security/pull/408#discussion_r460820020) discussion too. Thank you, @joshuagl, @lumjjb, @TheFoxAtWork !

PR #410, opened 2020-08-02
Add in-toto and harbor

ultrasaurus:
It looks like these became CNCF projects since original charter was written


ultrasaurus:
hmm...looks like Dragonfly and Harbor are under [SIG-Runtime](https://github.com/cncf/sig-runtime)

PR #540, opened 2021-02-19
Update permission for chairs & TLs

ultrasaurus:
this was apparently not updated a while back, we'll see if we can use settings instead of Admin UI


ashutosh-narkar:
LGTM

ashutosh-narkar (approved):
LGTM

TheFoxAtWork (approved):
LGTM

PR #541, opened 2021-02-19
add new TLs as repo CODEOWNERS

ultrasaurus:
before they were just added to README, but we want them to be the whole repo as is reflected in the comment


ashutosh-narkar (approved):
lgtm

lumjjb (approved):
LGTM

PR #542, opened 2021-02-19
add comment to explain better how this works

ultrasaurus:
we shouldn't use Github UI anymore because that is confusing


pragashj (approved):
LGTM

PR #555, opened 2021-03-04
self-assessment format tweak, added link to CII

jlk:
* Wrapped lines in md table aren't rendering right
* Added link to CII as that might not be a widely used acronym


TheFoxAtWork (approved):
LGTM

Issue #613, opened 2021-05-05
[Project] Convert sig-security to tag-security / Security TAG


TheFoxAtWork:
Description: In light of the TOCs approval to change SIGs to TAGs we need to update our naming convention in the repo

Impact: fairly large?  several of our  links are embedded as URLs not repo paths

Scope: figure out naming conventions, make a list of all the things we need to change outside of the repo, make the changes -- will take a bit of iteration from at least a few people who know the workings of the group to effectively brainstorm the list of things, new folks could be involved to help make the changes once the list of things to change is assembled  

TO DO
- [x] SIG Leadership: @ultrasaurus 
- [x] design team to make a new logo - in progress
- [x] Slide templates for kubecon & other presentations
- [X] Meeting notes google doc changes (thx @TheFoxAtWork)
- [X] Slack channel renames (Thanks @amye :) )
- [x] figure out appropriate short-names and edit [CONTRIBUTING.md#writing-style](https://github.com/cncf/tag-security/blob/main/CONTRIBUTING.md#writing-style)
- [x] make PR(s) with changes to repo
- [x] anything else?

lumjjb:
+ Require design team to make a new logo
+ Slide templates for kubecon

amye:
> * Require design team to make a new logo

Already in progress:) 
> * Slide templates for kubecon
> * Meeting note templates

Meeting Doc has been updated.
> * Slack channel renames (Thanks @amye :) )

ultrasaurus:
converted to a project, since it's already in progress and we def need to do it

PR #614, opened 2021-05-05
Changing SIG to TAG

amye:


.github/settings.yml
```diff
@@ -2,10 +2,10 @@ repository:
   # See https://developer.github.com/v3/repos/#edit for all available settings.
 
   # The name of the repository. Changing this will rename the repository
-  name: sig-security
+  name: tag-security
 
   # A short description of the repository that will show up on GitHub
-  description: 🔐CNCF Special Interest Group on Security -- secure access, policy control, privacy, auditing, explainability and more!
+  description: 🔐CNCF Technical Advisory Group on Security -- secure access, policy control, privacy, auditing, explainability and more!
```
TheFoxAtWork, line 8:
```suggestion
  description: 🔐CNCF Technical Advisory Group on Security (STAG) -- secure access, policy control, privacy, auditing, explainability and more!
```

TheFoxAtWork (approved):
It is important to note that any links in the repo that are URL not repo path may be broken.  issue #613 identifies this potential break

lumjjb:
It may be sufficient to just change links going forward since github does have an inherent redirection for renaming repos as long as a new repo isn't created in its place (i believe)

amye:
I just tested this with another TAG and it resolved correctly, so I think? we should be ok.

lumjjb (approved):
lgtm

pragashj (commented):
LGTM

PR #627, opened 2021-05-14
What: provide SSCSP to the repo

TheFoxAtWork:
Why:
* Press release is coming soon and users will be directed to the repo to download and read the PDF or the MD version


lumjjb (approved):
LGTM

PR #633, opened 2021-05-17
Add version commit hash for future diffing

lumjjb:
Added version commit hash so that when the paper needs to be updated, changes can be easily diff'ed to create an errata for the original version


TheFoxAtWork (approved):
<3 love this.  LGTM!

PR #674, opened 2021-05-31
add detail about how different github issue templates are used

ultrasaurus:
in talking about triage last week, how the different templates are used wasn't clear,
so I thought I would document how we had originally set that up


ultrasaurus:
@TheFoxAtWork @lumjjb this came up in talking about triage last week, so l would like your review!  

and, of course, anyone from the community, pls chime in if you have thoughts about our process

TheFoxAtWork (approved):
LGTM

PR #704, opened 2021-06-14
fix(ci): resolve issue with relative URL link check erroring

JohnHillegass:
Fixes an error when relative URL link check fails like in #701 

This config will replace relative URLs with absolute ones based off of main to better check the validity of the links


JohnHillegass:
@lumjjb FYI I saw an error with link checking in #701 on some of the relative URLs, this should fix for most cases

JohnHillegass:
Superceded by https://github.com/cncf/tag-security/pull/705

PR #709, opened 2021-06-20
Fix bug in spelling.sh

lumjjb:
@JohnHillegass can you verify that this is right? If not it seems like spell check stalls on xargs waiting on STDIN ?


ci/spelling.sh
```diff
@@ -2,4 +2,4 @@
 
 npm install -g cspell
 git fetch origin main:main
-git diff --name-only main $HEAD | grep .md | grep . && xargs -L1 npx cspell -c ./ci/spelling-config.json || echo "no markdown files changed"
\ No newline at end of file
+git diff --name-only main $HEAD | grep .md | grep . | xargs -L1 npx cspell -c ./ci/spelling-config.json || echo "no markdown files changed"
```
JohnHillegass, line 5: @lumjjb whoops I think I rushed this the other day and had only checked the negative case of "no markdown files changed". Lets add `--no-run-if-empty` flag instead to handle the case when no markdown files were changed.

```suggestion
git diff --name-only main $HEAD | grep .md | grep . | xargs --no-run-if-empty -L1 npx cspell -c ./ci/spelling-config.json
```

lumjjb:
@JohnHillegass made the suggested change.

JohnHillegass (approved):
Lgtm!

TheFoxAtWork (approved):
LGTM

PR #711, opened 2021-06-21
removing obsolete file

ultrasaurus:
can't recall why this yaml file was here, but all of its contents are obsolete, so this PR removes it


lumjjb (approved):
lgtm

Issue #687, opened 2021-06-04
TAG Security logo with transparent background doesn't look great in dark browser


jlk:
When browsing the TAG Security repo and viewing the main README.md in a browser running in "dark mode," the background color is dark, resulting in the TAG's logo being hard to read (screenshot below).

@lumjjb mentioned GitHub's tracking a potential fix for this at https://github.community/t/support-theme-context-for-images-in-light-vs-dark-mode/147981/23. Doesn't seem to be making progress, though...

<img width="930" alt="Screen Shot 2021-06-03 at 10 11 21 AM" src="https://user-images.githubusercontent.com/146853/120867638-2f30a880-c547-11eb-8442-d1882a6eb3f0.png">

PR #717, opened 2021-06-25
What: Uploaded more logos, corrected table, etc.

TheFoxAtWork:
Why:

* We were missing logos in the design readme
* the logo for the main readme was not darkmode friendly #687

This change addresses the need by:

* Created new logos
* Updated the table
* added the new dark mode safe logo to the main readme.


ultrasaurus:
fyi -- to disable the spell-check (where appropriate) you can add a comment above wherever the unique words are (if you don't want to add to the repo-wide config, which would likely be the case here):
```
<!-- cSpell:disable -->
```

ultrasaurus:
also, prolly even better, would be to ignore specific words



though common acronyms like "APAC" should prolly go in the repo-wide config `ci/spelling-config.json`

ultrasaurus:
oh, I just realized ALL of the errors are in the main README -- will try to work on separate PR to fix that!

ultrasaurus:
updated branch, spell check passes -- yay!

ultrasaurus:
dark mode fixed, here's what the logo in the README looks like this PR:
image

design/README.md
```diff
@@ -29,23 +29,14 @@
...
-        <td><img src="logo/cloud-native-security-horizontal-black.svg" width="200"></td>
-        <td><img src="logo/cloud-native-security-stacked-black.svg" width="95"></td>
-        <td><img src="logo/cloud-native-security-icon-black.svg" width="95"></td>
-</tr>
 <tr>
         <th>white</th>
-        <td><img src="logo/cloud-native-security-horizontal-white.png" width="200"></td>
-        <td><img src="logo/cloud-native-security-stacked-white.png" width="95"></td>
-        <td><img src="logo/cloud-native-security-icon-white.png" width="95"></td>
+        <td><img src="logo/Cloud-Native-Security-Horizontal-White.png" width="200"></td>
```
ultrasaurus, line 34: Please leave filenames consistent where possible.  Also, our style guide preferences all lower case filenames.

ultrasaurus:
White logos should be displayed on grey background.  This is what it looks like for me:
image

ultrasaurus (changes requested):
This is a great step forward, just a few small fixes and it'll be ready to merge.  See prior comments.

ultrasaurus:
on spellcheck, I didn't wait long enough, I guess... weird that it showed passing for a minute.  

I would suggest adding darkmodesafe to `/ci/spelling-config.json` since it shows up in a couple of files, but also ok IMHO if you want to use cspell:ignore locally in both files.

design/README.md
TheFoxAtWork: fixing.

TheFoxAtWork:
pretty-logos

TheFoxAtWork:
@ultrasaurus every change you requested should be resolved.  even added a preview of the new change!

ci/spelling-config.json
```diff
@@ -17,6 +17,7 @@
         "HIPAA",
         "CODEOWNERS",
         "explainability",
-        "APAC"
+        "APAC",
+        "darkmodesafe"
```
lumjjb, line 21: i think the better fix for this is for cpell to ignore that line instead.

```suggestion
        "APAC",
```

README.md
```diff
@@ -1,7 +1,6 @@
 # CNCF Security Technical Advisory Group
 
-![Cloud Native Security
-logo](design/logo/cloud-native-security-horizontal-color.svg)
+<img src="design/logo/cloud-native-security-horizontal-darkmodesafe.svg" alt="Cloud Native Security logo" width="400"/>
```
lumjjb, line 3:
```suggestion
<!-- cspell:disable-next-line -->
<img src="design/logo/cloud-native-security-horizontal-darkmodesafe.svg" alt="Cloud Native Security logo" width="400"/>
```

lumjjb (commented):
Looks good, trying another cspell fix

lumjjb:
@ultrasaurus `<!-- cspell:disable-next-line -->` seems to work

TheFoxAtWork:
@ultrasaurus can you verify the changes were applied so this can be merged?

lumjjb:
Linked this PR to cspell issue https://github.com/cncf/tag-security/issues/724 to address, if not should be good to merge... @ultrasaurus if you can take a quick look and we can merge this PR

ultrasaurus:
filenames should be all lower case:
image

not sure if that's the reason or if there are additional issues that cause broken images links: 
image

TheFoxAtWork:
@ultrasaurus looks like i've finally fixed it, it wasn't accepting the file rename which was why the images were broken.

design/README.md
```diff
@@ -1,17 +1,18 @@
-## TAG Security Logos
+## Security TAG Logos
 
-*Note: GitHub Flavored Markdown used in the Readme doesn't support background colors. The white logos below are displayed on the light grey of tables.*
+*Note: GitHub Flavored Markdown used in the Readme doesn't support background colors. The white logos below are displayed on the medium grey (#474756) in its row.*
```
ultrasaurus, line 3: Actually, it's not picking up that style.  It seems to be using a lighter grey
image

when I look at using the browser inspector, I see...
image

It is using the value `#f6f8fa`
image

ultrasaurus (commented):
so close... not quite sure what the issue is, but dropping in a comment with what I'm seeing

ultrasaurus:
@TheFoxAtWork looks like css is not supported by github, removed as part of the sanitization process: https://stackoverflow.com/questions/51956361/custom-css-file-for-readme-md-in-a-github-repo

I wonder if we could use the grey color swatch image as a background?  if github supports markup to overlay images or have background image or something

ultrasaurus:
.... OR was the note at the top of the file supposed to indicate that it looks incorrect in github, but would look correct rendered as HTML in a normal browser... if so, maybe it would be easier to just include explicit HTML, I dunno.  Confusing.

TheFoxAtWork:
@ultrasaurus i think i'll just render separate images just for the readme display with links to the source file without a background.

TheFoxAtWork:
Screen Shot 2021-08-05 at 10 05 36 AM
okay - i've modified the table for the white logos so that they 'appear' to be displayed in the readme on the correct background with a link to take the user to the correct file with a transparent background.

ultrasaurus (approved):
Sorry for the delay.  Looks fine.  I don't love that it'll be harder to maintain, but reasonable trade-off.  In retrospect, the initial version against light gray wasn't so bad, and I think that would be ok too.

One note -- didn't check dark-mode, where black prolly has same problem, but I don't think that needs to block this merging in.

PR #720, opened 2021-06-30
word wrap, formatting, spelling and broken links

ultrasaurus:
want to make sure this passes the linter before I do content edits!

pre-condition for: https://github.com/cncf/tag-security/issues/613

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -103,7 +103,7 @@ and provide concurrence prior to merging.
 Two persons with access, other than the PR author, must review the PR and
 provide concurrence, the last of which should perform the merge.
 
-[Code owners](CODEOWNERS.md) need to be at least one concurring reviewer or the
+[Code owners](CODEOWNERS) need to be at least one concurring reviewer or the
 merging party.
 
 ### Merging pull requests
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,6 +1,3 @@
----
-# cSpell:ignore Opps,usecase
----
 # Contributing
 
 We aspire to create a welcoming environment for collaboration on this project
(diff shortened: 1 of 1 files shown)


ultrasaurus:
I'm confused why linter would show these spelling errors...
image

@JohnHillegass @lumjjb

JohnHillegass:
Hmm @ultrasaurus I'll need to see if there's a work around to enable that but essentially anything that isn't in its dictionary is going to be flagged. I'm not sure if usecase or opps is proper spelling. We could put alternative spellings in the config file if it makes sense.

qnetter:
What is “opps”? Also, I’ve generally seen “use case” as two words.

ultrasaurus:
Opps is second word of a CamelCase name and usecase is within a link -- added inline comments for cspell to ignore these instances

CONTRIBUTING.md
```diff
@@ -47,24 +50,32 @@ language or language that is a matter of preferred usage are not considered
 nits.
 
 An example of preferential language:
```
lumjjb, line 49: Not quite related to the direct changes of this PR, but i think while we're at it, it may be easier to read if we have subheaders for preferential language vs nits
```suggestion
##### An example of preferential language:
```

CONTRIBUTING.md
```diff
@@ -47,24 +50,32 @@ language or language that is a matter of preferred usage are not considered
...
-```
+"Ability" is a human oriented term, "capability" is more technical and may be
+more appropriate.
+
+Suggestion:
+> They use cloud technologies with clear understanding of risks and the
+> capability to validate their security policy decisions are reflected in
+> deployed software.
 
 An example of a nit:
```
lumjjb, line 62:
```suggestion
##### An example of a nit:
```

lumjjb (approved):
LGTM! Thanks @ultrasaurus , couple nit suggestions

Issue #75, opened 2018-09-05
Request Briefing from Mastercard


ckemper67:
At Kubecon Mastercard expressed interest in Google's thoughts on OPA and we assume that Mastercard may be interested in ABAC. I think it would be interesting to hear from them in the context of SAFE. 

@dshaw mentioned that he has contacts at Mastercard.

lumjjb:
@dshaw any leads here or shall we close this in the meantime?

lumjjb:
@dshaw any leads here or shall we close this in the meantime?

lumjjb:
@dshaw shall we close this? and re-open if the conversation comes up again.

TheFoxAtWork:
closing stale and likely OBE (overcome by events)

PR #721, opened 2021-07-01
add new words to the spelling config

JohnHillegass:
@ultrasaurus this should address those issues you were seeing. Cspell splits camel case words and checks them against the dictionary so its not seeing opps and usecases in it then flagging them. This file extends the dictionary for cases like this. The entries are case insensitive so "opps" should cover "Opps".

Frees up #75

### DIFF
--- a/ci/spelling-config.json
+++ b/ci/spelling-config.json
@@ -26,6 +26,8 @@
         "Cappos",
         "ericavonb",
         "HIPAA",
-        "CODEOWNERS"
+        "CODEOWNERS",
+        "opps",
+        "usecase"
     ]
 }


ultrasaurus:
doesn't this generally allow "usecase" and "opps" as correct spellings?  is there any way to solve this more specifically?  like include the whole link or whole word or at least scope the special spelling to a particular file?  

it's too bad we can't add comments to the spelling config, cuz its json

JohnHillegass:
Closing in favor of inline spelling ignores

`<!-- cSpell:ignore Opps,usecase -->`

Issue #311, opened 2019-12-18
[Suggestion] markdown linting tools


kapilt:
Description: Research markdown linting tool for sig markdown files and prs.

- [ ] Link checking
- [ ] Linting (80 column chars)

others?

follow up from sig security meeting (dec 19th)

ultrasaurus:
added some common inconsistencies from [Contributor's Guide / Writing Style](https://github.com/cncf/sig-security/blob/master/CONTRIBUTING.md#writing-style) section

pbnj:
Is there agreement on markdown style, e.g. SETEXT vs ATX, CommonMark vs GitHub-Flavored Markdown?

Different tools have varying levels of support for different the different markdown specs and it would help narrow down the tool selection/evaluation process.

Some markdown linting tools I have come across in the past:

- https://github.com/remarkjs/remark-lint
- https://github.com/markdownlint/markdownlint
- https://github.com/DavidAnson/markdownlint
- https://github.com/textlint/textlint

A tool I have been happily using is: [prettier](https://prettier.io)

Pros:
- It's primarily used for formatting JavaScript, but it has built-in support for Markdown, as well as others, like YAML, JSON ...etc.
- The [rules/options](https://prettier.io/docs/en/options.html) are configurable, e.g. Print Width, Tab Width, Prose Wrap, End of Line ...etc.
- There are plenty of [editor integrations](https://prettier.io/docs/en/editors.html) across popular editors, e.g. atom, vscode, vim, emacs, and sublime.

Cons:
- It's a formatter, not a linter. So, running the tool will auto-correct issues instead of give you pointers to them. There is no way to have make it behave like a linter (AFAIK).
- It's not primarily a Markdown tool, thus it doesn't have a lot of rules specific to Markdown (e.g. header capitalization, spell-checking).

kapilt:
@pbnj thanks, i suspect we're targeting github flavored markdown, given the predominant usage of that style within the repo.

There's a dizzying plethora of tools in this space,  the two most common markdownlint (ruby & nodejs) don't appear to target the spelling or link checking use cases. it maybe we just need separate tools around each major use case (lint, link-check, spelling)

- https://github.com/lukeapage/node-markdown-spellcheck
- https://github.com/tcort/markdown-link-check

prettier seems interesting, afaics running as a code formatter is a different use case, but setup as a repo pre-commit hook would help automate some of the styling drudgery if its documented as part of the repo workflow (or perhaps a makefile target).

[...]

PR #723, opened 2021-07-01
fix to make README pass spellcheck lint

ultrasaurus:
* disable cSpell for the lists of names
* remove names from spelling config file
* rename `NEWMEMBER.md` => `NEW-MEMBER.md` (Aassuming no one is super attached to that name, this removes a cSpell exception)
* fixed error with spelling lint check when file is renamed


ultrasaurus:
request code review from @JohnHillegass and/or @lumjjb on the change to `ci/spelling.sh`

JohnHillegass:
On nice yep this lgtm! Hadn't considered impact of renaming files before.

NEW-MEMBERS.md
```diff
@@ -0,0 +1,29 @@
...
+New members are advised to:
+
+* Join the [CNCF Slack team](https://slack.cncf.io/), particularly
+  [#tag-security](https://cloud-native.slack.com/messages/CDJ7MLT8S) channel and
+  introduce yourself.
+* Initially go through the following documents in the repository:
+    * [README.md](https://github.com/cncf/tag-security/blob/main/README.md)
+    * [CODE-OF-CONDUCT.md][coc]
+    * [Use cases and personas][use-cases]
+* Regularly join one of the [Zoom meetings][meeting-time] at least for the first
```
ashutosh-narkar, line 16: Nit: `meeting-times`

NEW-MEMBERS.md
```diff
@@ -0,0 +1,29 @@
...
+  couple of months to get yourself up to speed.
+* Here are multiple ways to get involved:
+    * Join the meeting as advised above and express your area of interests or if
+      you want to work on any specific issue.
+    * Express your thoughts or ask questions on an issue you find interesting.
+    * Choose an issue where [help is
+      needed](https://github.com/cncf/tag-security/labels/help%20wanted) and
+      comment on it expressing interest.
+
+[meeting-time]: https://github.com/cncf/tag-security/blob/main/README.md#meeting-time
```
ashutosh-narkar, line 26: Nit: `meeting-times`

ci/spelling-config.json
```diff
@@ -4,28 +4,19 @@
     "ignorePaths": [
         "**/.github/**",
         "**/ci/**"
-    ],    
+    ],
     "words": [
```
ashutosh-narkar, line 8: Nit: We should add a comment explaining what goes in this list.

NEW-MEMBERS.md
ultrasaurus: not a nit!  that was an error -- the header changed, so did the anchor link.  will fix -- thanks!

ci/spelling-config.json
ultrasaurus: yeah, that would be nice.  Though a bit outside the scope of this change, since json doesn't support comments in any straightforward manner (that I know of). I made a note in the lint issue (which is still in-progress) https://github.com/cncf/tag-security/issues/311#issuecomment-872676561 -- feel free to add your thoughts there.

NEW-MEMBERS.md
ultrasaurus: fixed!

NEW-MEMBERS.md
```diff
@@ -0,0 +1,29 @@
...
+* Here are multiple ways to get involved:
+    * Join the meeting as advised above and express your area of interests or if
+      you want to work on any specific issue.
+    * Express your thoughts or ask questions on an issue you find interesting.
+    * Choose an issue where [help is
+      needed](https://github.com/cncf/tag-security/labels/help%20wanted) and
+      comment on it expressing interest.
+
+[meeting-times]: https://github.com/cncf/tag-security/blob/main/README.md#meeting-times
+[coc]: https://github.com/cncf/tag-security/blob/main/CODE-OF-CONDUCT.md
```
php-coder, line 27: nit:
```suggestion
[CoC]: https://github.com/cncf/tag-security/blob/main/CODE-OF-CONDUCT.md
```

NEW-MEMBERS.md
```diff
@@ -0,0 +1,29 @@
...
+know how you will contribute. The first step is to get yourself familiar with
+our mission at [Security TAG charter](governance/charter.md).
+
+New members are advised to:
+
+* Join the [CNCF Slack team](https://slack.cncf.io/), particularly
+  [#tag-security](https://cloud-native.slack.com/messages/CDJ7MLT8S) channel and
+  introduce yourself.
+* Initially go through the following documents in the repository:
+    * [README.md](https://github.com/cncf/tag-security/blob/main/README.md)
```
php-coder, line 13: Here and below: I'd suggest to use relative links rather than absolute. They are shorter and will work even after renaming a repo. For example:
```suggestion
    * [README.md](/blob/main/README.md)
```
But, please, double check that it works :)

NEW-MEMBERS.md
ultrasaurus: why capitalize reference links?  that's not visible to the reader and (for me) it makes it easier if all the "variable" are all lower case, so I don't have to remember what case was used if I want to use the reference elsewhere (unlikely in this file, but more a question of practice)

NEW-MEMBERS.md
ultrasaurus: `/README.md` works -- agree that is a good idea!  I just renamed the file, but I'll go ahead and make that change while I've got it open in my editor :)

NEW-MEMBERS.md
php-coder: Yes, you're right that isn't needed as it isn't visible to users. Disregard this comment then.

NEW-MEMBERS.md
ultrasaurus: fixed

php-coder (approved):
While I'm not part of tag-security, this is LGTM.

NEW-MEMBERS.md
```diff
@@ -0,0 +1,29 @@
...
+    * [CODE-OF-CONDUCT.md][coc]
+    * [Use cases and personas][use-cases]
+* Regularly join one of the [Zoom meetings][meeting-times] at least for the first
+  couple of months to get yourself up to speed.
+* Here are multiple ways to get involved:
+    * Join the meeting as advised above and express your area of interests or if
+      you want to work on any specific issue.
+    * Express your thoughts or ask questions on an issue you find interesting.
+    * Choose an issue where [help is
+      needed](https://github.com/cncf/tag-security/labels/help%20wanted) and
```
php-coder, line 23: `/labels/help%20wanted` didn't work or this just was overlooked?

ultrasaurus: I don't think that works because it's actually relative to something like `https://github.com/cncf/tag-security/blob/...` and I think it might be a bit fragile to use `../..`

ultrasaurus:
thanks @php-coder for your contributions!  You too could become a member of the group if you like: https://github.com/cncf/tag-security/blob/main/README.md#new-members

ashutosh-narkar (approved):
LGTM

PR #727, opened 2021-07-02
[WIP -- this is a test] 609 proposal project lint

ultrasaurus:


ci/spelling-config.json
```diff
@@ -4,28 +4,19 @@
     "ignorePaths": [
         "**/.github/**",
```
JohnHillegass, line 5:
```suggestion
        "**/.github/workflows/**",
```

JohnHillegass:
So I think where this is messing up is we excluded that whole `.github` dir from cspell execution but grep is seeing markdown files changed there and passing forward to xargs. Since the first arg is not empty xargs is taking it and passing to cspell. Cspell since its been told to skip that path, returns a non-zero code and xargs throws the 123. Changing like above handles this case. Whew spell checking is fun 😆

PR #729, opened 2021-07-06
README: Update timezone links to more helpful ones

axelsimon:
- Use "today" as a date, rather than a given date (should keep working even after DST changes)
- Show conversion only for the time we our interested in (the meeting) rather than every hour of the day
Note : for the APAC meeting, AEDT (Australia Eastern Daylight Time) is not currently observed, as East Australia is currently on AEST. Not sure what the best option is here, maybe clarify if we settle on Sydney time, for instance?

Also, change GMT to UTC, which is technically more correct (for what that matters…)


TheFoxAtWork (commented):
LGTM - this should be merged _after_ #730 to avoid the linter catching the same issues.

TheFoxAtWork (approved):
LGTM

PR #750, opened 2021-08-11
Add SLSA Maintainers to related groups

mlieberman85:
The above name of the group might change as the groups just
started and things like name, governance, etc. might change over
time. Will update if/when it does.


lumjjb (approved):
LGTM, will fix spellcheck/links in another PR

PR #767, opened 2021-08-28
Update CNSMap Service Meshes

MMerzinger:
Adds Open Service Mesh, updates Service Mesh Interface and removes Trivy (as it is no Service Mesh).


ashutosh-narkar (approved):
LGTM

PR #773, opened 2021-08-31
Minor fixes and improvements for CNSMap

MMerzinger:
Fixed typos and other minor things. Removed content in wrong place.

### DIFF
--- a/content/industries.md
+++ b/content/industries.md
@@ -40,19 +40,3 @@ public interest. In the public sector, it is essential to maintain harmony and t
 entities. Additionally, timeliness of deployments and features may also be a strong consideration. The adoption of
 cloud native, along with modern methodologies, can increase organizational velocity, which is critical to many areas of
 the public sector.
-
-## Projects
-- [Project Calico](https://www.projectcalico.org/)
-- [SPIFFE/SPIRE](https://github.com/spiffe)
-- [OPA](https://github.com/open-policy-agent/opa)
-- [Istio](https://github.com/istio/istio)
-- [Linkerd](https://github.com/linkerd)
-- [HyperLedger](https://www.hyperledger.org/)
-- [HashiCorp Vault](https://github.com/hashicorp/vault)
-- [Keylime](https://keylime.dev/)
-- [Kyverno](https://github.com/kyverno/kyverno)
-- [Keycloak](https://github.com/keycloak/keycloak)
-
-## Examples
-- Adopt a Zero trust architecture with least-privilege security policies. See [this](zero-trust-architecture#examples)
-  for more examples.
--- a/content/security-policies.md
+++ b/content/security-policies.md
@@ -4,14 +4,14 @@ date: "2020-10-01"
 category: "develop"
 ---
 
-It is essential to consider the security features and various configuration options of your orchestrator to control the security privileges the container runtime can use to spawn containers. The use of higher level policy and governance constructs may enforce those security guardrails).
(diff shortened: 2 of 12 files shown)


content/security-policies.md
```diff
@@ -4,14 +4,15 @@ date: "2020-10-01"
...
+It is essential to consider the security features and various configuration options of your orchestrator to control the security privileges the container runtime can use to spawn containers. The use of higher level policy and governance constructs may enforce those security guardrails.
 
 
 ## Projects
 - [OPA](https://www.openpolicyagent.org/)
 - [Kubernetes Admission Controller](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/)
 - [Kubernetes RBAC](https://kubernetes.io/docs/reference/access-authn-authz/rbac/)
-- [Kubernetes Pod Security Policy](https://kubernetes.io/docs/concepts/policy/pod-security-policy/) 
+- [Kubernetes Pod Security Policy (PSP) - (deprecated, removed in 1.25)](https://kubernetes.io/docs/concepts/policy/pod-security-policy/) 
+- [Kubernetes PSP Replacement Policy (alpha in 1.22)](https://github.com/kubernetes/enhancements/issues/2579)
```
ashutosh-narkar, line 15: Nit: We probably don't need to include this right now. Once there is an implementation we can add this in. WDYT ?

ashutosh-narkar (commented):
Thanks for this contribution @MMerzinger! Just one minor comment, rest lgtm.

content/security-policies.md
MMerzinger: I agree that it is too early for PSP replacement. Perhaps the reference to the deprecation of PSP would be useful, as the target group of readers may not know the stage of the lifecycle. But I have removed it with a new commit as it is clearly visible when you go to the link for PSP.

MMerzinger:
FYI: I added the sigstore and Connaisseur projects under Image Trust & Content Protection.

MMerzinger:
@ashutosh-narkar I fixed the merge conflicts and added myself to the contributors. Is the PR good for you or is there anything else to fix/change?

content/industries.md
```diff
@@ -40,19 +40,3 @@ public interest. In the public sector, it is essential to maintain harmony and t
...
-- [Linkerd](https://github.com/linkerd)
-- [HyperLedger](https://www.hyperledger.org/)
-- [HashiCorp Vault](https://github.com/hashicorp/vault)
-- [Keylime](https://keylime.dev/)
-- [Kyverno](https://github.com/kyverno/kyverno)
-- [Keycloak](https://github.com/keycloak/keycloak)
-
-## Examples
-- Adopt a Zero trust architecture with least-privilege security policies. See [this](zero-trust-architecture#examples)
-  for more examples.
```
ashutosh-narkar, line 58: Just for my understanding, why did you remove this ?

MMerzinger: From the perspective of the reader, each industry has its challenges and these challenges needs to be solved with e.g. a CNCF project. But currently there is no link between an industry and a project, hence it was confusing finding a list of projects that is almost a one to one copy from zero-trust-architectures. So I believe there is not "just" zero trust, that is important for those industries but even more topics and these are all covered by other chapters.

ashutosh-narkar: Thanks for the clarification.

ashutosh-narkar:
@MMerzinger can you please squash your commits and then we can merge this.

MMerzinger:
@ashutosh-narkar Okay like this or shall I squash the last two commits as well?

ashutosh-narkar (approved):

PR #789, opened 2021-09-08
Add buildpacks to assessment list and sort table alphabetically

anvega:


TheFoxAtWork (approved):
Merging even with the bad URL.  can't find a better one.

PR #812, opened 2021-10-22
Add klow / klown / okhsa, ua-parser-js, coa, rc breaches

adityasaky:
I'm a little unclear which compromise-definition applies to the klow/klown/okhsa situation, since they seem to have been new packages. I've added them separately because of the connection to ua-parser-js.


TheFoxAtWork (approved):
LGTM

PR #786, opened 2021-09-08
Cloud Custodian Joint Review markdown

sunstonesecure-robert:
needs review and final content

PR #819, opened 2021-11-18
C7n self assessment document

castrojo:
This is the Cloud Custodian self-assessment document that is needed as part of: https://github.com/cncf/tag-security/pull/786


castrojo:
Closing this as the content has been integrated into #786

PR #841, opened 2022-01-08
Update README.md

bobychaudhary:
Add Boby as a new CNCF member


lumjjb:
We no longer maintain a member list! However, we encourage you to still participate and get your name in on papers/deliverables and on the contributors list in github!

PR #842, opened 2022-01-10
What: Update the spelling config for linter

TheFoxAtWork:
Why:

* tech is very wordy and the linter doesnt speak tech without our help

This change addresses the need by:

* increasing the spelling config based on c7n PR


ashutosh-narkar (approved):
Not sure if we can use some regex in the `words` key. Otherwise the list keeps growing.

TheFoxAtWork:
> Not sure if we can use some regex in the `words` key. Otherwise the list keeps growing.

No idea. Problem for another day?

PR #883, opened 2022-04-04
Update remove APAC meeting (no longer running)

lumjjb:


TheFoxAtWork (approved):
LGTM - would recommend providing a way for folks interested in setting up a region specific meeting to do so with the leadership team so we can support them.

lumjjb:
Created https://github.com/cncf/tag-security/issues/886 to track your suggestion!

JustinCappos (approved):
Agreed we should drop this.    

I wish we could have gotten more momentum here...  :(

PushkarJ:
Should we wait to merge this until we can close out on https://github.com/cncf/tag-security/issues/886 ?

lumjjb:
don't think we have to wait since we do have a tracking issue and don't want folks to try to dial in to the meeting

PR #907, opened 2022-05-19
Secure Software Factory Reference Architecture markdown and pdf

anvega:


lumjjb:
There are some critical changes in the pdf that are not applied yet. We can
merge just the markdown for now.

PR #910, opened 2022-05-19
Update README - Fix PDF v2 wording

Boes-man:
I believe it should be "Posted PDF second version", not "Posted PDF first version"?


TheFoxAtWork (approved):
LGTM

PR #916, opened 2022-05-30
docs: adding reference article

lirantal:
Adding another reference article with regards to the Colors and Faker npm modules incident


lirantal:
The build is broken due to cspell but not on any spelling issues, but rather on the fact it can't run due to unsupported Node.js version of the workflow. Looking into it...

lumjjb (approved):
LGTM!

lumjjb:
Looks good - seems like there's a DCO check enabled now as well, would you be able to add a sign off to your commits? 

Yea - we will have to take a look at the cspell issue - would you mind raising an issue for this?

lumjjb:
re: cspell - may be a flake actually, after fixing the DCO let's see if it comes up again..

lirantal:
Yep, I'll work out the signed commits later in this PR and the others, no worries.
Thanks for quickly taking a look at it!

PR #918, opened 2022-05-30
build: use Node.js LTS for tooling

lirantal:
Explicitly set a supported Node.js version which some tooling on this workflow requires.
In fact, CI currently fails on the `Spelling` step due to defaulting to Node.js 12, yet requiring Node.js 14 ([see here](https://github.com/cncf/tag-security/runs/6649092902?check_suite_focus=true)), such as:

```
+ cspell@6.0.0
added 150 packages from 65 contributors in 8.822s
From https://github.com/cncf/tag-security
 * [new branch]      main       -> main
 * [new branch]      main       -> origin/main
ApplicationError: Unsupported NodeJS version (12.22.12); >=14 is required
```

PR #924, opened 2022-06-02
Upgrade node:12 to node:18

PushkarJ:
What: This PR upgrades node images in `docker-compose.yml` file to node:18 from node:12. It also reverts the change made in #918 

Why: `cspell` was failing earlier for all PRs. This PR attempted to fix it #918 but was unsuccessful since the node version used is defined `docker-compose.yml` file.


TheFoxAtWork (approved):
LGTM !

PR #932, opened 2022-06-11
Update tech-lead-transition.md

lumjjb:
Too much privileges in the group and it's not really used. The group is given admin.

Signed-off-by: Brandon Lum <lumjjb@gmail.com>


TheFoxAtWork (approved):
LGTM

PR #945, opened 2022-06-26
docs: add pypi malicious packages incident

lirantal:
Signed-off-by: Liran Tal <liran.tal@gmail.com>

### DIFF
--- a/supply-chain-security/compromises/2022/pypi-malicious-packages.md
+++ b/supply-chain-security/compromises/2022/pypi-malicious-packages.md
@@ -1,0 +1,49 @@
+<!-- cspell:ignore pyg-modules -->
+<!-- cspell:ignore pygrata -->
+<!-- cspell:ignore pygrata-utils -->
+<!-- cspell:ignore loglib -->
+<!-- cspell:ignore loglib-modules -->
+<!-- cspell:ignore hkg-sol-utils -->
+<!-- cspell:ignore hxxp -->
+<!-- cspell:ignore Backdoored -->
+
+# PyPI python packages caught sending stolen AWS keys to unsecured sites
+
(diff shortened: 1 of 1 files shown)


SantiagoTorres:
Hi @lirantal ! thanks for the submission. Could you help me with:

1. specifying if this was a typosquat. If this wasn't a typosquat, could we figure out how they got access to these packages?
2. Searching for perhaps a monetary impact, or identify a high profile victim for this. 

The latter is mostly because it helps people understand the effect of the compromise

lirantal:
@SantiagoTorres:
1. There are relevant named modules on PyPI (like `loglib`, and `modules`, but there isn't something wholeistic like `loglib-modules`) so in a sense, this isn't your ordinary typosquatting.
2. No big impact that I'm aware of, this was part of a pentester's "let me show you how bad things can go". Other than that I'm not aware of any high profile victims due to this incident. We might discover about this in the future since some of the data was leaked on a public endpoint accessible to anyone to harvest.

SantiagoTorres:
Ahhh, I see now. Could you please add that context? it's not exactly a typosquat, but rather.... ehh, a "typophish" or something of the like. I would assume the names wereen't chosen randomly, but rather to make sure people would select these as they provide things deemed useful.

As for impact, I think we can also mention that (i.e., this was an exercise, real impact not measured and such)

lirantal:
@SantiagoTorres I don't know what the term "typophish" is and I want to refrain from inventing new ones and confusing readers :-)

That said, I've updated the description of the attack to be typosquatting in nature, and mentioned that this was a "proof of concept" employed attack, however it did end up causing data exposure. 

Ready on my end now with DCO ✅

SantiagoTorres (approved):

PR #986, opened 2022-10-03
Adding an example to "primitive workloads that rely on the orchestrator"

paolomainardi:
I am not sure to have understood it, so I am going to propose this change to validate it.

Signed-off-by: Paolo Mainardi <paolo.mainardi@sparkfabrik.com>


PushkarJ:
@paolomainardi container runtimes like containerd, cri-o are good examples. Specifically in case of kubernetes, kubelet API is also a valid example here.

anvega (approved):

PR #990, opened 2022-10-18
Adding Intel Alder Lake BIOS leak

Caze121:
New compromise regarding source code leak


lumjjb (approved):
LGTM! Missing DCO, should be good otherwise!

PR #1007, opened 2022-11-14
[[Compromises]] Trusting Trust

zerb4t:


lumjjb:
Probably should add the compromise type as [Dev tooling](https://github.com/cncf/tag-security/blob/main/supply-chain-security/compromises/compromise-definitions.md#dev-tooling). In addition, can you put the links in the references as well. Thanks!

szh (approved):
This is great! I love how this is from before most of us were born 😆

mnm678 (approved):

PR #1010, opened 2022-11-25
[compromises] npm API timing attack

zerb4t:


lumjjb:
It seems like this is more a weakness than a comrpomise. would like to hear others’ thoughts on this as well. @mlieberman85 @mnm678

mnm678:
> It seems like this is more a weakness than a comrpomise. would like to hear others’ thoughts on this as well. @mlieberman85 @mnm678

I think the loss of privacy of package names could be a compromise, though not a high impact one on its own. Knowledge of these package names could be used for a further attack like dependency confusion.

szh:
Agreed. I'd vote to close

PR #1036, opened 2023-02-15
Replace "artefact" with "artifact" in documents & small linting fixes

Vombato:
# Description

Reading the documents I've seen many times referencing "artifact" as "artefact", knowing that both versions of the word are correct in common English, I think standardizing the word in all the project would be beneficial, also given the fact that in many external references is mostly used the "artifact" version.
That being said, I think that this will clear some confusion for non-native English speakers approaching these subjects for the first time.

## Type of change

- [x] Replaced all references of "artefact" with "artifact"
- [x] Fixed some markdown linting of bullet lists and paragraphs (also cleared some trailing spaces)

# Checklist:
- [x] My code follows the style guidelines of this project

Thanks,
Luca Bandini


Vombato:
> See relevant Slack thread: https://cloud-native.slack.com/archives/CDJ7MLT8S/p1666724370786239

Sorry I missed that, I've just joined that channel.
It still sounds a bit odd to me to use the project lead’s regionality and therefore "artefact", when all external references use another term.
If that's the final decision then I apologize 🙌🏻

szh:
Personally I also think it makes sense to standardize on one spelling. Maybe one of the tech leads can chime in on this.

mlieberman85:
I wonder if some of this can be resolved through internationalization (i18n). I think it's probably fine to have working documents use spelling based on those leading/contributing the bulk, but it is causing confusion on the finished rendered documents. Given that these spellings can easily be switched with locale changes in things like Google docs it might just be worthwhile to render the pdfs in both US and UK English?

anvega:
We won't be making changes to the existing markdown but instead opt for regenerating pdf assets in both US and UK English. Tracked by https://github.com/cncf/tag-security/issues/1052

Issue #994, opened 2022-10-25
Auto Request Review Action is failing


PushkarJ:
Unrelated to changes in the PR, but looks like auto-request-review job fails in this PR: https://github.com/cncf/tag-security/pull/993 

Best guess for the reason is that we missed adding @lirantal in this file as a person in "triage" or "read" role: https://github.com/cncf/tag-security/blob/main/.github/settings.yml

If we add them as a collaborator, this should most likely be fixed. More info on other available permissions:

Available permissions:

> From least access to most access, the roles for an organization repository are:
> 
> Read: Recommended for non-code contributors who want to view or discuss your project
> Triage: Recommended for contributors who need to proactively manage issues and pull requests without write access
> Write: Recommended for contributors who actively push to your project
> Maintain: Recommended for project managers who need to manage the repository without access to sensitive or destructive actions
> Admin: Recommended for people who need full access to the project, including sensitive and destructive actions like managing security or deleting a repository

Ref: https://docs.github.com/en/organizations/managing-user-access-to-your-organizations-repositories/repository-roles-for-an-organization#repository-roles-for-organizations

szh:
Also appears to be affecting #1020

PushkarJ:
@mnm678 can you help approve this fix: https://github.com/cncf/tag-security/pull/1021 ?

PR #1053, opened 2023-04-04
Fix typo NP --> npm

joshuagl:


szh (approved):
LGTM, nice catch

mnm678:
Thanks for the pr! The failing check is unrelated and tracked in #994

PR #1084, opened 2023-06-21
Update README.md

anvega:
Enlarges logo to full width

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 <!-- cspell:disable -->
 <!-- markdownlint-disable-next-line MD033 MD013 -->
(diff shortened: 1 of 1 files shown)


JustinCappos:
I would think we might read the room to see if this is desirable (if this hasn't already happened).  It feels a little overly large to me now, but I'm fine with whatever others like!

anvega:
There was an adjustment period from going small to large in repos like https://github.com/envoyproxy/envoy and https://github.com/spiffe/spire, but it achieves a better visual for newcomers. Looks more like a site and less like a document.

PushkarJ (approved):

PR #1086, opened 2023-06-23
Add repojacking example from 2021.

evankanderson:
I was going through the catalog of supply chain compromises, and noticed that a novel (to me) publication from 2021 wasn't included.


anvega (approved):

anvega:
Thanks, @evankanderson!

Issue #554, opened 2021-03-03
Security Pal for Self-Assessments Pilot


TheFoxAtWork:
Description: Execute a pilot that introduces and encourages one or two projects to complete a [self-assessment](https://github.com/cncf/sig-security/blob/master/assessments/guide/self-assessment.md) with a CNCF SIG-Security person assigned to walk them through and guide them on completing the self-assessment.
It is not intended to be a joint-evaluation, more to be as a touch point with security insights to guide projects in jump starting their security considerations for their software project.

Ideally we would select one or two projects that are not security focused (don't provide security in the ecosystem) that are very early in their maturity, sandbox preferred, to test this out with.  At the end of this we want to capture:

* The usefulness of the self-assessment by the project
* the value of having a short-term Security-minded pal/buddy/advisor
  * specific to the completion of the self-assessment
  * in general

ContainerD has a [similar concept](https://github.com/containerd/project/blob/master/GOVERNANCE.md#security-advisors) however it is more for day-to-day security advisement

Impact: We believe this will be immensely helpful for early maturity, non-security projects to begin thinking about and considering their security.  The product of this also is an input to the [joint-review](https://github.com/cncf/sig-security/blob/master/assessments/guide/joint-review.md) under the [Security Review Process](https://github.com/cncf/sig-security/tree/master/assessments/guide).

Scope: We would need to research and select one or two projects, determine if they are amenable to trying this out, and then have them complete the self-assessment.  Overall the assessment completion is relatively light-weight, and should take no more than a few hours, with overall commitment from the SIG-member to be about a week or two.

TO DO
- [X] SIG Leadership Representative - @TheFoxAtWork 
- [X] Project lead
  * @jlk
- [X] Security 'pal'
  * @magnologan 
  * @jlk 
  * @apmarshall

jlk:
I'm a fan of getting security stuffs outside our bubble to non-infosec peeps, so count me in. :)

apmarshall:
Likewise, happy to help out however I can with this.

magnologan:
Same here, but my vote still goes to Security Champions for the name of this program! =)

jlk:
Breaking things down in more detail, I guess what we need to cover...

1. Research what ContainerD has learned (all of us)
1. Pick say 3-4 projects from https://www.cncf.io/sandbox-projects/ (looking for 23 but we'll get blowoffs) (all)
1. Each reach out to 1-2 projects to see if we can get their interest 
[...]

PR #1094, opened 2023-07-12
Security Pals retrospective slides

jlk:
Adding a markdown version of the powerpoint deck I used at a STAG meeting to give a retrospective of the "Security PALS" pilot documented in #554.

I've written this for [Marp presentation system](https://marp.app/) which seems to render OK as plan markdown, but if folks think that's overkill, I'll remove the extra markup.


TheFoxAtWork:
Thank you so much for pulling this together!! (It's been a while but I'm thrilled this seems fruitful!)

I'd love for security pals to engage with accepted sandbox security projects (minimally) as they are under the TAG's domain.  I suspect plenty of other projects would also gain value from this - not just sandbox.  Outreach would certainly be a good start, but also allowing projects to come to the TAG and ask for assistance.  These were some of the original intentions around this.

JustinCappos:
@jlk @TheFoxAtWork There will be a meeting on Monday, July 31st at 1PM Eastern about how Security Pals, Security assessments, Lightweight threat modeling should be combined.  ( https://zoom.us/j/99063331512?pwd=WlFFMGhWM2FhcGtDaVJJSm5peDk5Zz09 ).  Please feel free to invite anyone who has an opinion on this topic.

TheFoxAtWork:
Thanks Justin!  I won't be able to attend, however be sure to reflect this in the monthly TAG update to the TOC with your liaisons' awareness.

PushkarJ:
@JustinCappos do we have any suggestions for @jlk for this PR, based on the outcomes of the meeting held on July 31? 

cc @sublimino

JustinCappos:
@jlk Based upon what we learned from your experience and our discussion, we are going in a different direction with some aspects of this.  From what we've decided, there is overlap, but enough differences that I'm a little concerned that having these slides as is will be confusing. 

I will close this for now, but am open to discussion about how to make modifications to accept this.  However, I don't think it's a good idea to take it as it is now.

PR #1106, opened 2023-09-08
Update README.md

anvega:
Update meeting ID. 

The motivation for this change, automated livestream and upload to youtube broke a while back with changes to how livestream is configured. It no longer allows use of streaming using personal room id.

Also retitled US and EMEA to US and Western Hemisphere and Eastern Hemisphere for inclusivity.

PR #1109, opened 2023-09-11
fix: update Zoom link

sublimino:
- this change moves the Zoom link to the group's room
- the group's room supports auto-streaming to YouTube

I've updated the two meeting Calendar Invites on CNCF TAG-Security Public Calendar to reflect this change.


PushkarJ (approved):
Approving this as discussed to fix streaming. Naming discussion can continue in https://github.com/cncf/tag-security/pull/1106

PR #1114, opened 2023-09-26
i18n EMEA meeting time from London to UTC

anvega:
LF guidance is to internationalize the schedule by displaying meeting times in UTC and to decouple from centralizing to the local timezone of a particular location. 

* Replaces London for UTC+0 and UTC+1 to account for the difference in daylight savings during summer months.
* Replaced time.is link from what time is in London to what time is UTC.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,8 +78,8 @@ Group meeting times are listed below:
 
 - US:   Weekly on Wednesdays at 10:00am UTC-7 (see your timezone
   [here](https://time.is/1000_today_in_PT?CNCF_Security_TAG_US_Meeting))
-- EMEA: Bi-weekly on Wednesdays at 01:00pm London (see your timezone
-  [here](https://time.is/1300_today_in_London?CNCF_Security_TAG_EMEA_Meeting))
+- EMEA: Bi-weekly on Wednesdays at 12:00pm UTC+0/01:00pm UTC+1 (see your timezone
+  [here](https://time.is/UTC?CNCF_Security_TAG_EMEA_Meeting))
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -78,8 +78,8 @@ Group meeting times are listed below:
 
 - US:   Weekly on Wednesdays at 10:00am UTC-7 (see your timezone
   [here](https://time.is/1000_today_in_PT?CNCF_Security_TAG_US_Meeting))
-- EMEA: Bi-weekly on Wednesdays at 01:00pm London (see your timezone
-  [here](https://time.is/1300_today_in_London?CNCF_Security_TAG_EMEA_Meeting))
+- EMEA: Bi-weekly on Wednesdays at 12:00pm UTC+0/01:00pm UTC+1 (see your timezone
```
mnm678, line 81: Can we make it clear when it's UTC+0 vs UTC+1? Otherwise LGTM

anvega: Yes. Added explanation describing its UTC+1 during daylight savings.

mnm678 (approved):

sublimino:
Hi team, I'm on PTO this week but this change removes daylight savings for EMEA in the meeting time link — which is observed in the EMEA meeting times and so will be incorrect. 

The new wording here is also unnecessarily confounding. UTC is not observed in the UK. We use GMT or BST, which are equivalent to PST and PDT.

https://time.is/GMT-BST might be a better link to be equivalent to "PT", but I'd suggest reverting this PR and understanding requirements first.

The TOC [doesn't observe this purported guidance](https://github.com/cncf/toc#meeting-time) (and [links to](https://dateful.com/time-zone-converter?t=8:00AM&tz=San%20Francisco) San Franciso time) and it's unclear why these PRs to EMEA naming and timing are being made without attendee-driven requirements or prior art — one for us to discuss @anvega.

TOC:
image

PR #1115, opened 2023-09-27
Add unconference template

mnm678:
Add a template for coordinating unconference talks at the Security Hub!


mlieberman85 (approved):
LGTM, we just probably also want to have some docs that link to this template. Also we should have a basic explanation of an unconference for folks to read through.

Issue #1116, opened 2023-09-29
TAG Security website info is wrong


jberkus:
This page: https://tag-security.cncf.io/

With source here: https://github.com/cncf/tag-security/blob/main/website/content/_index.md

... has the info for TAG-App Delivery, not TAG-Security.  Please update!

ragashreeshekar:
I have created a service ticket for the same. I'll follow up and relay here as I receive updates.
[CNCFSD-1988](https://cncfservicedesk.atlassian.net/servicedesk/customer/portal/1/CNCFSD-1988)

jberkus:
Er, shouldn't you just create a PR?  You own that repo.

PushkarJ:
@jberkus yes you are right, the website content is co-owned by us and CNCF staff (see the commit history). We are planning to discuss how to make changing this easier in the future as part of that ticket so PRs would be the way to go.

ragashreeshekar:
Received updates on the service ticket, and I have created the PR to address this issue. 
Reference - https://github.com/cncf/tag-security/pull/1122.

anvega:
Thanks @jberkus. We've fixed this but looks like the page has yet to be regenerated.

PushkarJ:
This is now fixed post merge of https://github.com/cncf/tag-security/pull/1150

PR #1122, opened 2023-10-11
Static update to the TAG Website landing page with contents from TAG README.md

ragashreeshekar:
Per. CNCF service ticket https://cncfservicedesk.atlassian.net/servicedesk/customer/portal/1/CNCFSD-1988, TAG website template is a starting point and we maintain it henceforth. As discussed with the Chairs + TLs updating our website landing page with contents from our TAG README.md. 

This is a static update to address https://github.com/cncf/tag-security/issues/1116, however we might need to brainstorm approaches to ensure the README.md is mirrored in the website index for long term sustenance efforts.

### DIFF
--- a/website/content/_index.md
+++ b/website/content/_index.md
@@ -15,36 +15,234 @@ toc_hide: true
 </div>
 
 
-The TAG produces guidance for and gathers feedback from cloud app users and
-developers and provides guidance and coordination to CNCF projects in the TAG's
-technical domains.
+# CNCF Security Technical Advisory Group
 
-- [TAG Charter](https://github.com/cncf/toc/blob/main/tags/security.md)
-- [Community events](https://community.cncf.io/tag-app-delivery/)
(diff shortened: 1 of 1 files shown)


anvega (changes requested):
I noticed a few areas in the README that could benefit from updates:

* Workgroups Listing: The current list seems to not include all active workgroups. It would be beneficial for all teams if we could ensure every active workgroup is acknowledged in the README.

* Chairs Listing: This section also appears to be out of date. Let's make sure we reflect the current leadership accurately.

* Under the 'security reviews' section, my affiliation needs an update. I'd appreciate a quick revision there.

The considered mirror push strategy ensures content uniformity but introduces a delay between the README updates and the website content while the mirror push review happens at the destination. I think I would be beneficial to see that implementation first in the form of a chained pr before making this change in isolation risking the mirroring not happening a later time. 

That said, given the website's information changes less frequently than the README, perhaps we could reconsider this approach. Would it make sense to directly adjust a few lines on the website repo occasionally for clarity and accuracy instead of relying solely on the mirror push?

PR #1146, opened 2023-11-16
Merge "Open and Secure" TM and Assessments Book Release 

anvega:
Adds final proofed copy of the assessments book and removes early release copy.


JonZeolla:
@JustinCappos @anvega Is there a md version of this as well?  I would be happy g doc -> md if that's not done, assuming the g doc content is aligned to the pdf

anvega:
@JonZeolla  No immediate plans to maintain a public markdown version. For now, we intend for the PDF to be a means of public consumption with maybe an epub sometime soon. There are some workflow complexities in publishing which having a third asset to keep in sync would lead to drift between multiple versions. I do appreciate the offer.

mnm678 (approved):

PR #1213, opened 2024-01-05
chore: Add eddie-knight as reviewer for assessments

eddie-knight:


JustinCappos:
@sublimino This would be a great thing to merge sooner rather than later so we can close the loop on these.

PushkarJ (approved):
Approving on behalf of @sublimino since it's end of the day for him.

PR #1217, opened 2024-01-18
change from facilitators to facilitator for security assessments

anvega:
Stepping down from my position as one of the designated facilitator for the TAG security assessments after evaluating my current commitments and responsibilities, both professionally and personally.

The process is currently very effective with the time and energy that Justin dedicates to it, and I'm eager to see how the TAG security assessments continue to grow and evolve.

I look forward to continuing to contribute to the TAG in the capacity of TL.


JustinCappos (approved):
Sorry to see you go, but I understand.  Would be happy to see someone step up in the future.

Issue #1221, opened 2024-01-23
GH actions are always failing to Git safe directory error, but returning success


tuminoid:
Hello tag security maintainers! 

I randomly noticed that your Github actions are always failing to Git safe directory check, but returning success, ie. false positives.

In the `Makefile` you have targets that are called like `make links`, and then in `ci/links.sh` you want to iterate over changed files via `for file_name in $(git diff --name-only $HEAD main); do`  but git returns nothing on stdout due safe directory failure (all output goes to stderr), so you end up linting nothing in any of your actions. `set -e` is not enabled for the scripts, so it just falls thru.

For example, on PR #1202 [linter action logs](https://github.com/cncf/tag-security/actions/runs/7559235882/job/20582614675?pr=1202)

```console
Run make lint
Running lint...


Creating network "tag-security_default" with the default driver
Creating tag-security_lint_run ... 
Creating tag-security_lint_run ... done

added [5](https://github.com/cncf/tag-security/actions/runs/7559235882/job/20582614675?pr=1202#step:3:6)9 packages in 3s

1[8](https://github.com/cncf/tag-security/actions/runs/7559235882/job/20582614675?pr=1202#step:3:9) packages are looking for funding
  run `npm fund` for details
npm notice 
npm notice New minor version of npm available! [10](https://github.com/cncf/tag-security/actions/runs/7559235882/job/20582614675?pr=1202#step:3:11).2.3 -> 10.3.0
npm notice Changelog: <https://github.com/npm/cli/releases/tag/v10.3.0>
npm notice Run `npm install -g npm@10.3.0` to update!
npm notice 
fatal: detected dubious ownership in repository at '/usr/src/app'
To add an exception for this directory, call:

	git config --global --add safe.directory /usr/src/app
warning: Not a git repository. Use --no-index to compare two paths outside a working tree
usage: git diff --no-index [<options>] <path> <path>
...
```

Same for [spell check](https://github.com/cncf/tag-security/actions/runs/7559115852/job/20582205190?pr=1201), [links check](https://github.com/cncf/tag-security/actions/runs/7559115852/job/20582204878?pr=1201) and even the [setup job](https://github.com/cncf/tag-security/actions/runs/7559115852/job/20582196771?pr=1201).

If you run the linters locally over the repository content as suggested by the `ci/*.sh` scripts, you get:
- about 10000+ markdownlinter complaints
- 10000+ spelling errors (not including failing every word on the spanish whitepaper....)
- 460 broken links

[...]

PR #1223, opened 2024-01-25
ci: fix ci linter scripts

tuminoid:
Linting scripts were failing to Git safe dirertory check, both when run via GH actions, or locally via docker-compose. To fix the scripts, few things need to be changed.

1. Add /usr/src/app to Git safe directories, so Git will run any Git commands, such as "git fetch ..." in that directory.
2. Add "set -e" so command failure will fail the script, to produce correct return code for the check.
3. Due "set -e", and the loop we are running, we need to ignore the linting command returning failure, and capture the failure to be returned later command.
4. We also need to remove the grep from the pipeline. If the commit has no matching files, it would return false error, thanks to "-o pipefail". Same can be applied to all scripts, having Git filter the wanted files, instead of having additional if clause.

Git safe directory check was introduced already in Git 2.35.2 in 2022, and the "ubuntu-latest" image used by the GH action shouldn't have taken  too long to upgrade to that or newer Git version. The old GH action logs only go back a month, and they're all showing the same. It could  be that the actions have been broken for 1.5 years. The same can bereproduced locally, as the node:18 image used by the docker-compose is also having recent enough Git.

This is tested locally to produce correct behavior for all three cases, and also when no matching files have been changed.

Fixes: #1221

### DIFF
--- a/ci/links.sh
+++ b/ci/links.sh
@@ -1,10 +1,16 @@
 #!/usr/bin/env bash
 
+set -e
+shopt -s globstar
+
+FAILURE=0
+
+git config --global --add safe.directory /usr/src/app
 npm install -g markdown-link-check
 git fetch origin main:main
 # To run this on the entire repo, replace the following command with `$(find ./ -type f | grep .md)`
-for file_name in $(git diff --name-only $HEAD main); do
-  if [[ $file_name == *".md" ]]; then
-    npx markdown-link-check --config ./ci/link-config.json --progress --verbose "$file_name"
-  fi
+for file_name in $(git diff --name-only $HEAD main -- ./**/*.md); do
+  npx markdown-link-check --config ./ci/link-config.json --progress --verbose "$file_name" || FAILURE=1
 done
+
+exit $FAILURE
--- a/ci/spelling.sh
+++ b/ci/spelling.sh
@@ -1,7 +1,13 @@
 #!/usr/bin/env bash
 
+set -eo pipefail
+shopt -s globstar
+
 function spellcheck () {
-    git diff --name-only --diff-filter=AM main $HEAD | grep '\.md$' | xargs --no-run-if-empty -L1 npx cspell --show-suggestions -c ./ci/spelling-config.json
+    FAILURE=0
+    git diff --name-only --diff-filter=AM main $HEAD -- ./**/*.md | \
+        xargs --no-run-if-empty -L1 npx cspell --show-suggestions -c ./ci/spelling-config.json || FAILURE=1
+    return $FAILURE
 }
 
 function printhelp () {
@@ -21,6 +27,7 @@ function printhelp () {
 }
 
 
+git config --global --add safe.directory /usr/src/app
 npm install -g cspell
 git fetch origin main:main
 # Print help if spellcheck fails
--- a/ci/lint.sh
+++ b/ci/lint.sh
@@ -1,10 +1,16 @@
 #!/usr/bin/env bash
 
+set -e
+shopt -s globstar
+
+FAILURE=0
+
+git config --global --add safe.directory /usr/src/app
 npm install -g markdownlint-cli
 git fetch origin main:main
 # To run this on the entire repo, replace the following command with `$(find ./ -type f | grep .md)`
-for file_name in $(git diff --name-only $HEAD main); do
-  if [[ $file_name == *".md" ]]; then
-    markdownlint -c ./ci/lint-config.json "$file_name"
-  fi
+for file_name in $(git diff --name-only $HEAD main -- ./**/*.md); do
+  markdownlint -c ./ci/lint-config.json "$file_name" || FAILURE=1
 done
+
+exit $FAILURE
--- a/ci/setup.sh
+++ b/ci/setup.sh
@@ -1,5 +1,8 @@
 #!/usr/bin/env bash
 
+set -e
+
+git config --global --add safe.directory /usr/src/app
 git fetch origin main:main
 git diff --name-only main $HEAD > modified
 echo "these files have changed"


PushkarJ:
Thank you so much for opening the issue and PR to fix this. 

One request, When you ran this locally can you please paste the output for the jobs in a comment? I want to see if we can fix those spellchecks and lints _before_ we merge this so CI Jobs don't fail on all open PRs.

tuminoid:
[tag-lint.txt](https://github.com/cncf/tag-security/files/14062076/tag-lint.txt)
[tag-spelling.txt](https://github.com/cncf/tag-security/files/14062081/tag-spelling.txt)
[tag-links.txt](https://github.com/cncf/tag-security/files/14062084/tag-links.txt)

eddie-knight:
Very excited to see this being caught and remedied. 

Is it prudent to use this spellchecker, considering the significant volume of false positives in the sample output? Seems like it'll be more overhead than its worth... This may be a good time to disable that, unless other folks are happy about using it.

tuminoid:
> Very excited to see this being caught and remedied.
> 
> Is it prudent to use this spellchecker, considering the significant volume of false positives in the sample output? Seems like it'll be more overhead than its worth... This may be a good time to disable that, unless other folks are happy about using it.

Having a spell checker is definitely a matter of opinion, especially adding (here, fixing it) to a repository with considerable content violating it.

We (Metal3) also recently started using cspell for our website repository, converting from yaspeller (sending everything to Yandex servers for spellchecking is not cool). It is worthwhile it do spend a bit of time analyzing where the failures are coming from. For example, there is a Spanish document and each word is obviously false negative, so a rule to ignore Spanish documents cuts down the number. Similarly, many hits come from author names, so ignoring a line where authors are listed cuts it more. We also had plenty of issues coming from code fences (triple backticks) and command quote (single backticks), which we now ignore. Combined with a project specific ignore list (you have a list going already), it wasn't too much effort in the end to get to zero spelling issues.

Also, by default, this repository only checks the changed files, so if the PRs are primarily adding new documents, they don't need to worry about cleaning the mess. Only providing clean new documents. This should be easily achievable if few rules are put in place to avoid false positives.

That said, I suggest you merge the fixes to the spell checker (this PR) regardless of the decision. This way you get a working local spell check at least. Then you can disable the spell checking for PRs by removing it from the GH action in separate PR.

ci/spelling.sh
```diff
@@ -1,7 +1,12 @@
 #!/usr/bin/env bash
 
+set -eo pipefail
+
 function spellcheck () {
-    git diff --name-only --diff-filter=AM main $HEAD | grep '\.md$' | xargs --no-run-if-empty -L1 npx cspell --show-suggestions -c ./ci/spelling-config.json
+    FAILURE=0
```
PushkarJ, line 6: Can we leave the markdown filtering as is ? the new filter does not seem to work when I tested it locally

```
personalmacbook:tag-security demo$ git diff --name-only --diff-filter=AM main $HEAD -- ./**/*.md ./*.md 
personalmacbook:tag-security demo$ git diff --name-only --diff-filter=AM main $HEAD
ci/links.sh
ci/lint.sh
ci/setup.sh
ci/spelling.sh
security-whitepaper/v2/cloud-native-security-whitepaper-it.md
security-whitepaper/v2/cloud-native-security-whitepaper-simplified-chinese.md
personalmacbook:tag-security demo$ git diff --name-only --diff-filter=AM main | grep '\.md$'
security-whitepaper/v2/cloud-native-security-whitepaper-it.md
security-whitepaper/v2/cloud-native-security-whitepaper-simplified-chinese.md
```

ci/spelling.sh
```diff
@@ -1,7 +1,12 @@
...
+    git diff --name-only --diff-filter=AM main $HEAD -- ./**/*.md ./*.md | \
+        xargs --no-run-if-empty -L1 npx cspell --show-suggestions -c ./ci/spelling-config.json || FAILURE=1
+    return $FAILURE
```
PushkarJ, line 9:
```suggestion
    FAILURE=0
    git diff --name-only --diff-filter=AM main $HEAD | grep '\.md$' | \
        xargs --no-run-if-empty -L1 npx cspell --show-suggestions -c ./ci/spelling-config.json || FAILURE=1
    return $FAILURE
```

Something like this works with `make spelling` to pick up the markdown files correctly

ci/links.sh
```diff
@@ -1,10 +1,15 @@
...
+
+git config --global --add safe.directory /usr/src/app
 npm install -g markdown-link-check
 git fetch origin main:main
 # To run this on the entire repo, replace the following command with `$(find ./ -type f | grep .md)`
-for file_name in $(git diff --name-only $HEAD main); do
-  if [[ $file_name == *".md" ]]; then
-    npx markdown-link-check --config ./ci/link-config.json --progress --verbose "$file_name"
-  fi
+for file_name in $(git diff --name-only $HEAD main -- ./**/*.md ./*.md); do
```
PushkarJ, line 11: Same request as spelling for linting as well.

ci/spelling.sh
PushkarJ: I also tested this with `make spelling` and new filtering did not pick the markdown files changed but the older filtering did.

PushkarJ (changes requested):
Thank you so much again @tuminoid for the PR. I noticed a small edit needed to markdown filtering, otherwise looks good. Let me know if I missed anything!

Additionally, I agree and confirm that only changed files will get flagged going forward. So this is okay to merge without fixing the whole repo before that.

As an aside, this brings up a good point about translations being the biggest cause of alerts that can be suppressed. We can fix that trivially by moving translations into a separate sub-directory under `v2` and `v1` and then ignoring that path here: https://github.com/cncf/tag-security/blob/main/ci/spelling-config.json#L4-L7 . This can definitely be done in a new PR that I hope to get one of our newer contributors to work on.

ci/spelling.sh
tuminoid: This is coming from the missing `shopt -s globstar` on your local shell (its responsible for expanding `**`) , which I have enabled on mine. I added it to the scripts, now it works as expected. Sorry about that.

ci/spelling.sh
tuminoid: LIke mentioned, the chained pipe is problematic for picking up the errors. This would actually produce a failure if case no markdown file was edited, like in this PR. If grep doesn't find files, it returns an error, which would be picked up as a spelling error.

ci/links.sh
tuminoid: See the first comment, it has now been fixed in the script.

tuminoid:
> Thank you so much again @tuminoid for the PR. I noticed a small edit needed to markdown filtering, otherwise looks good. Let me know if I missed anything!

Hi @PushkarJ ! Thanks for reviewing and testing. You actually caught an issue in the scripts, thanks for that. It is very old fashioned "works on my machine" case, as I have by default enabled `shopt -s globstar` on my default shell, and forgot to add it to the linter scripts as well. Globstar option is responsible for expanding `**` correctly in Bash, and as it wasn't enabled, it did not pick up any markdown files in subdirectories. 

I've fixed the issue in the linter scripts, and personally I'd like to stick with the new way. Mostly as it would be the same in all three scripts and that the possible pipefail in 'spelling.sh` would mess up the error picking logic. I will of course amend the PR the way you like, if you don't agree, but I'd argue this is cleaner and better way.

> Additionally, I agree and confirm that only changed files will get flagged going forward. So this is okay to merge without fixing the whole repo before that.

Just for the record: It has been the case all along, this PR did not chance that.

> As an aside, this brings up a good point about translations being the biggest cause of alerts that can be suppressed. We can fix that trivially by moving translations into a separate sub-directory under `v2` and `v1` and then ignoring that path here: https://github.com/cncf/tag-security/blob/main/ci/spelling-config.json#L4-L7 . This can definitely be done in a new PR that I hope to get one of our newer contributors to work on.

:+1:

PushkarJ:
Thank you @tuminoid. I can confirm your fix with globstar option is working so I am okay to keep this. 

Please rebase the branch to current main branch and then we can merge this 🚀

tuminoid:
> Please rebase the branch to current main branch and then we can merge this 🚀

Rebased!

PushkarJ (approved):

PR #1224, opened 2024-01-28
Various fixes to supply chain compromises catalog

adityasaky:
This PR makes various fixes to the catalog of supply chain compromises.

- Adds references in listing where missing
- Recategorizes some incidents based on whether source code repository was affected or just release artifacts


JustinCappos (approved):
Thanks for these updates.  LGTM

PR #1237, opened 2024-04-03
Swap in PageFind in place of Google Custom Search

cjyabraham:
From https://github.com/cncf/cncf.io/issues/836:

GCS doesn't seem to produce very good search results on small sites. It'd be better to remove our reliance on Google as well. Instead we can use [PageFind](https://pagefind.app/), which is an open source local search for static sites.


eddie-knight:
## Before:

Screenshot 2024-04-11 at 2 11 24 PM

## After:

Screenshot 2024-04-11 at 2 11 53 PM

sublimino (approved):
LGTM, thank you for the contribution :pray:  @eddie-knight will fix linting issues in #1241

PR #1241, opened 2024-04-11
Refining website to use primary repo content

eddie-knight:
Currently the website requires independent maintenance from the rest of the repo. This change will allow us to create pages that pull their content from other files in the repo at build time.


eddie-knight:
Links check is failing due to internal links on the markdown page table of contents, nothing we can do about that as far as I can tell.

anvega (approved):

PushkarJ (approved):

Issue #1257, opened 2024-05-22
Website Content Review


eddie-knight:
In order to increase the quality of outputs from TAG Security, to simplify the project maintenance, and to streamline new member familiarization, there is a need to do a large-scale cleanup of this repository. This will involve the revision or removal of files and/or directories.

As a first phase, we will be scoping focus specifically to content that is displayed on the website:

```sh
# snippet from website/Makefile
--include='assessments' --include='assessments/**' \
--include='governance' --include='governance/**' \
--include='supply-chain-security' --include='supply-chain-security/**' \
--include='*.md' --exclude='*'
```

This issue description will be used to summarize unique work items that we have found, which should be tackled. Unless otherwise noted, PRs and comments are welcome from anyone in the community to address the questions or problems outlined below.

## Action Items

```[tasklist]
### Assessments
- [ ] Add a draft title or issue reference here

### Events
- [ ] Add a draft title or issue reference here

### Governance
- [ ] Add a draft title or issue reference here

### Supply Chain Security
- [ ] Add a draft title or issue reference here

### Blog
- [ ] Add a draft title or issue reference here
```

PushkarJ:
Thank you for creating this space to track all this upcoming work. 

While we are working on this, let's see if we can add a banner "website under construction" 🚧 on website pages and link this issue in the banner for any feedback folks may have :)

mrcdb:
Great initiative @eddie-knight !

With regards to the point related to the TSSA package, I think clarification may be needed in the [self-assessment guide](https://github.com/cncf/tag-security/blob/main/assessments/guide/self-assessment.md#self-assessment-use) which currently states that:

> This document provides the CNCF TAG-Security with an initial understanding of [project] to assist in a joint-assessment, necessary for projects under incubation. Taken together, this document and the joint-assessment serve as a cornerstone for if and when [project] seeks graduation and is preparing for a security audit.

eddie-knight:
Thanks @mrcdb — I'll keep the list updated with any comments that roll in here.

eddie-knight:
Notes from one point of discussion on today's NA TAG meeting.

## Problem

Recent discussion has raised concern as to whether the`governance/` directory has a reasonable amount of clarity with regard to the intent of its contents. Similarly, it has been suggested that there are too many top-level files that relate to project governance. (for example, multiple files related to the project license).

## Proposed Solution

[...]

PR #1258, opened 2024-05-23
Added construction banner

eddie-knight:
Part of #1257 

[[Preview](https://deploy-preview-1258--tag-security.netlify.app)]

### DIFF
--- a/website/layouts/partials/under-construction-tmp.html
+++ b/website/layouts/partials/under-construction-tmp.html
@@ -8,8 +8,9 @@
 				This site may contain outdated or incomplete information.
 			</p>
 			<p class="m-0">
-				We aim to have all content updated before 
-				<a href="https://events.linuxfoundation.org/cloudnativesecuritycon-north-america/">June 26th</a>!
+				<a href="https://github.com/cncf/tag-security/issues/1257">
+					Follow along or join the revision effort!
+				</a>
 			</p>
 		</div>
 	</div>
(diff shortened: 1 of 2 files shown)


brandtkeller (approved):

website/layouts/partials/under-construction-tmp.html
```diff
@@ -0,0 +1,16 @@
...
+	<div class="row">
+		<div class="col-1">
+			<i class="fas fa-hammer"></i>
+		</div>
+		<div class="col-11">
+			<p class="m-0">
+				This site may contain outdated or incomplete information.
+			</p>
+			<p class="m-0">
+				We aim to have all content updated before
```
PushkarJ, line 11: I feel since this is opensource work, we can make this more generic to say in the coming weeks, instead of fixed deadline. We still also in my opinion think that we should add the website issue link to point folks who want to help, so that they know where to get started or show interest. Even if two folks organically join and show interest that's better than nothing :)

PushkarJ (commented):
Thank you!

website/layouts/partials/under-construction-tmp.html
eddie-knight: "Coming weeks" requires a point-in-time to give it meaning. I'll add a link to the GitHub issue per your suggestion... that should be sufficient to remove any time references.

PushkarJ (approved):

PR #1121, opened 2023-10-10
Added provenance-implementation hub w/ README and example

eddie-knight:
This is based on the proposal made by the TAG Security community to house provenance ingestion documentation for projects, following the approach that was requested by @mlieberman85 and @mnm678.

I've created a subdirectory with instructions and a single example (Argo CD).

PR #1268, opened 2024-06-06
Delete invalid CODEOWNERS file as we use roles

anvega:
Permissions and commit privileges in this repository are managed using roles. It appears that the codeowners file was introduced through https://github.com/cncf/tag-security/pull/1121. However, the current configuration contains errors, as the specified user does not have write access.


mnm678 (approved):
For context, the roles process is documented here: https://github.com/cncf/tag-security/blob/main/governance/github.md

PR #1275, opened 2024-06-12
docs: Update SSC compromise catalog

abhisek:
This is first in a (expected) series of PRs to update the supply chain compromises catalog. 

- Fix link in index
- Update incident title to make it verbose
- Add note to review incidents that are possibly outside the scope of SSC security

### DIFF
--- a/supply-chain-security/compromises/2010/fsf-website.md
+++ b/supply-chain-security/compromises/2010/fsf-website.md
@@ -1,5 +1,7 @@
 # Free Software Foundation Website Hack
 
+**Note:** Review if this incident can be categorized as supply chain incident as per [compromise definitions](../compromise-definitions.md)
+
 The source repository for the FSF's website was hacked via a SQL injection.
 
 ## Impact
--- a/supply-chain-security/compromises/2010/fsf-website.md
+++ b/supply-chain-security/compromises/2010/fsf-website.md
@@ -1,7 +1,5 @@
 # Free Software Foundation Website Hack
 
-**Note:** Review if this incident can be categorized as supply chain incident as per [compromise definitions](../compromise-definitions.md)
-
 The source repository for the FSF's website was hacked via a SQL injection.
 
 ## Impact
@@ -11,7 +9,7 @@ exfiltrate user names and encrypted passwords from the affected server.
 
 ## Type of Compromise
 
-Source Code
+Attack Chaining
 
 ## References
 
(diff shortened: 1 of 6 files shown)


supply-chain-security/compromises/2010/fsf-website.md
```diff
@@ -1,5 +1,7 @@
 # Free Software Foundation Website Hack
 
+**Note:** Review if this incident can be categorized as supply chain incident as per [compromise definitions](../compromise-definitions.md)
```
mnm678, line 3: This looks like it was categorized as a supply chain attack because the compromise happened at the source code

supply-chain-security/compromises/2013/apt.md
```diff
@@ -1,5 +1,7 @@
 # APT didn't enforce signature validation for source packages
 
+**Note:** Review if this incident can be categorized as supply chain incident as per [compromise definitions](../compromise-definitions.md)
```
mnm678, line 3: This looks to fall under "Trust and Signing"

mnm678 (commented):
Thanks for the changes! Let's see if we can get the categorization notes resolved before merging.

supply-chain-security/compromises/2013/apt.md
abhisek: I agree. The reason I flagged this is because we are defining this project as `compromise` catalog which would probably mean that a vulnerability was actually exploited in real life. In this case I see the issue as a vulnerability without any real life exploitation evidence or trigger for a breach / incident.

I am removing the note because I see the scope of this project is to catalog real life attack vectors that can cause supply chain issues and not just incidents.

supply-chain-security/compromises/2010/fsf-website.md
abhisek: Reading the referenced doc at https://www.computerworld.com/article/1439287/free-software-foundation-s-software-repository-hacked.html

I see original entrypoint for attacker was SQL Injection. Subsequently they extracted credentials from the database through SQL injection, uploaded a PHP reverse shell and gained access to the server that contains the source code for the website. There doesn't seem to be any evidence that this source code is in turn consumed by any other party. This seems like a first party attack without any impact to a third party.

If we still have to retain this, I feel we should change the classification type to `Attack Chaining` because a multi-step attack was performed to gain access to the server and deface the website.

mnm678: Re-categorizing this as Attack Chaining makes sense to me

abhisek: @mnm678 Updated the category. Over to you.

mnm678 (approved):

PR #1276, opened 2024-06-12
Moved contribution-related docs into CONTRIBUTING.md

eddie-knight:
Part of #1257, this aims to simplify the top-level and `governance/` directories by moving anything related to contribution into `CONTRIBUTING.md`


anvega (changes requested):
Thanks for your efforts in organizing the repository. I have a few thoughts regarding the proposed changes:

It's not idiomatic to have a CONTRIBUTING/ directory for contribution-related documentation. Typically, CONTRIBUTING.md is expected to be a single file, as outlined in the [GitHub documentation](https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/setting-guidelines-for-repository-contributors).

Some good examples of contribution guidelines are:

[GitHub Docs contribution guidelines](https://github.com/github/docs/blob/main/CONTRIBUTING.md).
[Ruby on Rails contribution guidelines](https://github.com/rails/rails/blob/main/CONTRIBUTING.md).
[Open Government contribution guidelines](https://github.com/opengovfoundation/open-gov-guide/blob/main/CONTRIBUTING.md).

I encourage merging the different files into a single CONTRIBUTING.md file instead of creating a directory. This approach is more in line with common repository structures and improves ease of use for contributors.

Additionally, I suggest including this change as part of the broader Information Architecture reorganization for the CNCF TAG Security Repository. You can find more details and participate in the discussion [here](https://docs.google.com/document/d/1bdcZ3WIBmLrTTZj_1aLp3fGl68WqEhifItIHThE53k8/edit).

eddie-knight:
The directory has already existed, not created by this PR. Will look at the rest of your feedback shortly.

abhisek:
I will agree with @anvega As someone looking to get started in contributing to TAG security, I was looking for `CONTRIBUTING.md` and totally missed the CONTRIBUTING directory till I was pointed out about resources in it.

eddie-knight:
The contributing dir has existed for the past couple years... Would like to hear from other chairs & TLs about whether this dir is a good candidate for removal. Happy to migrate all of the relevant material to a single md.

@mnm678 @PushkarJ WDYT?

mnm678:
There seems to be enough content for a directory, but maybe we can add a CONTRIBUTING.md that links to the files in the directory

eddie-knight:
@mnm678 I tried my hand at merging all of the content into a single file, and I think it works well enough. Check it out and let me know what you think.

anvega (approved):
Looks good!

CONTRIBUTING.md
```diff
@@ -0,0 +1,204 @@
...
+
+To become familiar with the team and understand how you will contribute, start
+by reading our [TAG Security charter].
+
+New members are advised to:
+
+- Join the [CNCF Slack], and introduce yourself on the the [#tag-security]
+  channel.
+- Review the following documents:
+  - [README.md]
```
brandtkeller, line 18: This link works for traversal in a local directory but is non-functional in the website.

eddie-knight: Oh this is frustrating. Do you see the different behavior between README and CODE-OF-CONDUCT?

The latter properly directs to `./CODE-OF-CONDUCT` while the former inexplicably goes to `contributing/README.md`.

eddie-knight: Troubleshooting... Confirmed they are both being parsed as local refs by `website/layouts/_default/_markup/render-link.html`

eddie-knight: Troubleshooting... In `render-link` README does not register as having a value for `$url.Fragment` while CODE-OF-CONDUCT does.

eddie-knight: Troubleshooting... The issue is because L7 of `renderlink` resets the value of $url, and if the file was a README it will be a null value at this point.

eddie-knight: Troubleshooting... L9-19 are responsible for removing the `contributing/` from the path. We need that cut in all situations, but for some reason we are hitting an edge case here.

eddie-knight: Since the problem was just with links to the top-level README, I added handling for that edge case specifically.

PR #1300, opened 2024-06-26
Removed CNCF Projects file as unnecessary and unmaintained.

eddie-knight:
Part of #1257, pairing with @brandtkeller

We believe this file is both unmaintained and not adding particular value.


governance/charter.md
```diff
@@ -39,7 +39,7 @@ improve overall security in cloud native systems.
 
 ## Focus
 
-In addition to the [CNCF security-related projects](cncf-projects.md), there
```
mnm678, line 42: We could replace this with a link to the CNCF landscape filtered to TAG Security (https://landscape.cncf.io/?group=projects-and-products&view-mode=grid&tag=security)

mnm678 (approved):

jkjell (approved):

PR #1307, opened 2024-07-03
Fix website publications table link column

mrcdb:
This SCSS fix addresses the following action item in https://github.com/cncf/tag-security/issues/1257

> The publications table is too wide— in the final column "Link" four letters is getting broken up into two lines.


eddie-knight (approved):
Very cool. 

Minor change, but I'll leave this open until the EMEA meeting later today.

Issue #398, opened 2020-06-10
[Suggestion] Simplify the Roles page


yeelinglam:
**Description**: provide an at-a-glance on all the roles currently available

**Impact**: clarify the list of roles, especially for prospect members

**Scope**: not yet determine and require discussions with the group

Additional Considerations:
- Add "Active Member" requirement section
- Tie it to current list of people serving in each role as a potential for shadowing for anyone who may want to take on a role in the future

anvega:
Closing this as we moved away from membership list and we now document official tag leadership roles.

szh:
Looks like https://github.com/cncf/tag-security/blob/main/governance/roles.md?plain=1#L44-L45 is out of date. Should it be removed? Or is there a different action that should be taken?

mnm678:
> Looks like https://github.com/cncf/tag-security/blob/main/governance/roles.md?plain=1#L44-L45 is out of date. Should it be removed? Or is there a different action that should be taken?

It looks like that is left over from the previous process and can be removed.

PR #1310, opened 2024-07-03
Remove outdated info about documenting members

szh:
As per [this comment](https://github.com/cncf/tag-security/issues/398#issuecomment-2206903201), the roles.md document still says that all members are listed in the repository, which is no longer true. This PR removes this outdated language.


eddie-knight:
Thanks @szh!

Do you mind if I push a commit to the PR branch with the linting fixes for this file? Or you can do it if you'd like, but we don't want to have your name with a red X beside it in the commit history. 

We're slowly but surely getting all the linting done across all the repo files!

szh:
Absolutely, go ahead!

eddie-knight (approved):

PR #1389, opened 2024-10-15
feat(assessments): Adds wasmCloud project self-assessment

thomastaylor312:
Adds the completed TAG Security self-assessment. Please let me know if we missed anything important!


JustinCappos (approved):
Looks great!   We'll have a few questions during the assessment process, but it is a great start!

PR #1417, opened 2024-12-04
Updating the Makefile script on the website to work natively with MacOS.

pedroignacio13:
After having some issues setting up the environment locally on MacOS, we figured it out what the problem was.
This PR adapts the Makefile script to work correctly when setting up the environment for local development.

Thanks Brandt and Eddie for the help on this issue!

### DIFF
--- a/website/Makefile
+++ b/website/Makefile
@@ -14,7 +14,7 @@ deps:
 
 	# Update all imported markdown files to work as standalone hugo pages (except READMEs, see below)
 	# sed command is configured for the Netlify ubuntu env
-	# If you're NOT running this on a mac, you'll need to remove the empty string after -i (line 25)
+	# If you're running this on a mac, you'll need to add double quotes after -i in the sed command below: `sed -i "" "1s/^/$$text_to_prepend/" "$$file"; \`
 	find root -type f -name '*.md' | while IFS= read -r file; do \
 		base_name=$$(basename "$$file" .md); \
 		if [ "$$file" = "/_index.md" -o "$$base_name" = "README" ]; then \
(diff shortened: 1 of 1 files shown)


mrcdb:
I am not sure if this is run in CI as well, wouldn't it have an impact as the "new" sed instruction would always run? I am wondering if there would be a better way of running the script for MacOS vs Linux (e.g. with a flag or something like that)

pedroignacio13:
I'll let this PR open and try to find alternatives to this.
Let me know if you have something on your mind.

brandtkeller (changes requested):
I agree with @mrcdb - this doesn't work with linux (and the expected CI) processes natively. 

Until such a time where this workflow is refactored - we may want to consider a conditional for macos/linux OR requiring gnu-sed on macos development environments.

website/Makefile
```diff
@@ -14,14 +14,15 @@ deps:
...
+	# If you're NOT running this on a mac, you'll need to remove the empty string after -i (line 25)
 	find root -type f -name '*.md' | while IFS= read -r file; do \
 		base_name=$$(basename "$$file" .md); \
 		if [ "$$file" = "/_index.md" -o "$$base_name" = "README" ]; then \
 			continue; \
 		fi; \
 		title_case=$$(echo "$$base_name" | sed -e 's/-/ /g' -e 's/\b\(.\)/\u\1/g' | sed 's/cncf/CNCF/Ig'); \
 		text_to_prepend="---\ntitle: \"$$title_case\"\n---\n"; \
-		sed -i "1s/^/$$text_to_prepend/" "$$file"; \
+		sed -i "" "1s/^/$$text_to_prepend/" "$$file"; \
```
eddie-knight, line 25:
```suggestion
		sed -i "1s/^/$$text_to_prepend/" "$$file"; \
```

eddie-knight (commented):
Suggested changes will modify this to just being a comment so that it has the necessary info, but still defaults to working for CI builds

@pedroignacio13

eddie-knight (approved):

eddie-knight:
@brandtkeller I've modified this to only contain the instructional comment for developer instruction, which is definitely needed.

brandtkeller (approved):

PR #1420, opened 2024-12-05
self-assessment: add Confidential Containers

mythi:
Add self-assessment of Confidential Containers project (CNCF Sandbox).


JustinCappos (approved):
It's a little short / terse in places, but this can be addressed if / when they go for a joint assessment.

mythi:
> It's a little short / terse in places, but this can be addressed if / when they go for a joint assessment.

@JustinCappos thanks for the quick review and approval! I wanted to clarify that I used links to your website docs on purpose because that's where we maintain most of the content these assessments are looking to see. This makes the maintenance easier.

PR #1427, opened 2024-12-12
Add champions to moving-levels-review-template.md

eddie-knight:


brandtkeller (approved):
There something about this I want to augment - but I am unsure what it is as of yet. Otherwise this looks good to me.

PR #1445, opened 2025-01-31
Update OpenTelemetry Security Self-Assessment

austinlparker:
As part of our graduation efforts (see https://github.com/cncf/toc/pull/1271), OpenTelemetry is submitting an updated security self-assessment for review.


JustinCappos (approved):
thanks for updating this.

PR #1450, opened 2025-03-03
Proposing a queue for cases in the community compromises catalog

zerb4t:
Sometimes security researchers and the media will discuss an incident and describe it as a supply chain attack. This is starting to happen faster than we can review them and evaluate how they fit the criteria for adding them to the catalog. The goal of this queue is to keep track of the cases. It also helps current and future contributors by giving them good first tasks. If someone reads the incident descriptions and determines it doesn't meet the criteria, they can delete the entry. Alternatively, if they write a short page about it, they can delete the entry when they make their PR. I'm looking forward to hear what others in the community think about this.

### DIFF
--- a/community/catalog/compromises/queue.yaml
+++ b/community/catalog/compromises/queue.yaml
@@ -1,0 +1,19 @@
+cases:
+  - urls:
+    - https://www.darkreading.com/threat-intelligence/iranian-apt-targets-us-drokbk-spyware-github
+    - https://www.secureworks.com/blog/drokbk-malware-uses-github-as-dead-drop-resolver
+    notes: log4j to vmware horizon path, use of github for c2
+  - urls:
+    - https://www.darkreading.com/cyberattacks-data-breaches/chinese-apt-targets-hong-kong-in-supply-chain-attack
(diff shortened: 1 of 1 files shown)


mnm678 (approved):

mnm678:
I like this idea! It should give an easy way to add links to new incidents, then allow others to write up the catalog entry.

I'll leave this a couple days for others to weigh in.

brandtkeller (approved):
Likewise appreciate the initiative with this. Covers a distinct gap.

brandtkeller:
(disregard the failing "Auto Request Review" check - that is not related to this PR)

eddie-knight:
@riaankleinhans Are you familiar with the pending+required check here? None of the STAG leads are grokking this at the moment.

`WIPExpected — Waiting for status to be reported`

eddie-knight:
I'm not sure what the origin of that WIP required test was... But since it wasn't doing anything, I've removed it from the list of required checks in the `main` branch protection settings.

riaankleinhans:
> @riaankleinhans Are you familiar with the pending+required check here? None of the STAG leads are grokking this at the moment.
> 
> `WIP Expected — Waiting for status to be reported`

No, not familiar.

PR #1452, opened 2025-03-05
Added latest PDF of the v2 whitepaper japanese translation

eddie-knight:
This is the latest version of the japanese translation PDF, provided by CNCF staff.

We will need a final review to ensure the content is generally readable and contains the expected content.


y-tabata:
This may be due to the font, but in Japanese, "、" and "。" are normally placed at the bottom, just like commas and periods in English, but in this white paper they are placed in the middle, which looks strange.

eddie-knight:
Thanks @y-tabata! I've reached out to CNCF to let them know

y-tabata:
There were some inappropriate Japanese translations of "authorization". I submitted a PR to fix the MD version, which is probably the basis for the PDF version.
https://github.com/cncf/tag-security/pull/1455

y-tabata:
@eddie-knight 
Thanks for merging. LGTM as my side.

PR #1476, opened 2025-07-01
self-assessment: add KubeVirt

jean-edouard:
Add self-assessment of KubeVirt


JustinCappos (approved):
Looks like a good start on this.   It wouldn't be a huge lift to turn this into the shape needed for a joint assessment here.   Mostly there would need to be some more clarity to help people understand how this is used in normal situations, etc.  From my quick look, most of the threat descriptions, etc. seem really nicely detailed!

evankanderson:
It looks like this got merged into the wrong place; #1479 to relocate it.

PR #1479, opened 2025-07-11
Move KubeVirt assessment to projects folder

evankanderson:
The KubeVirt self-assessment was merged to the wrong folder, which puts it alongside the `Projects` header on the website.


eddie-knight (approved):

eddie-knight:
Thanks for catching this!

PR #1480, opened 2025-07-13
Removed old+broken reviewer assignment action

eddie-knight:
I noticed that PR#1452 has been sitting unreviewed idle because it has a ❌ from a failing check that doesn't really matter. This PR removes that check.


evankanderson (approved):

mnm678 (approved):
We may want to revisit this in the future, but with the repository changes it will probably look different

PR #1499, opened 2025-11-04
fix(blog): move blog article to blog directory

brandtkeller:
Moving the security assessment blog to the blog directory so that it will be publicly viewable.


eddie-knight:
Thanks for getting to the bottom of this

PR #1507, opened 2025-12-04
Add redirect for PRs to appropriate GitHub repo

evankanderson:
Suggested by @JustinCappos, thanks!


brandtkeller (approved):
I think this is the appropriate foundation to build on. There might be some nuance for content transfer to the contribute (docs) repository but we can sort that out.

eddie-knight (approved):
excellent idea

